PR #5, opened 2015-08-10
Oom reporting

erichkeane:
Allow Out-Of-Memory size reporting


erichkeane:
Thiago: This passes all unit-tests, and works with the new one.  Most of this is per-our previous discussion.

thiagomacieira:
Merged.

PR #7, opened 2015-09-15
Add convenience function cbor_encode_text_stringz()

lpereira:
This calculates the length of a text string to be encoded using
strlen().  Useful when encoding constant strings.


thiagomacieira:
Thanks

PR #8, opened 2015-09-15
Fix conversion of large negative integers to string

thiagomacieira:
There was an off-by-one error due to the way we handled them.


thiagomacieira:
@erichkeane Can you take a quick look?

thiagomacieira:
Merged

PR #10, opened 2015-09-16
Print unsigned integer values suffixes with "u"

lpereira:


thiagomacieira:
Why? There is no ambiguity and clearly cborpretty can't be used for C code generation...

lpereira:
I am dumping messages received by IoTivity so I can encode a similar
message in another implementation. This helps me distinguish value types
without looking at the code that generated them.

thiagomacieira:
As I said, there's already no ambiguity. Floating point always has a dot, negative values have a minus sign. Anything left is unsigned. There's no need for "u".

lpereira:
I thought that, since there are functions to encode both signed and unsigned integers, there would be difference in the encoding. Reading the implementation and the RFC I stand corrected.

thiagomacieira:
The function is a convenience. It encodes positive int64_t as unsigned integers (major type 0) and negative as negative (major type 1).

PR #11, opened 2015-09-18
Micro-optimise cbor_value_get_{float,double}

thiagomacieira:
floats and double are too big for value->extra and the
CborIteratorFlag_IntegerValueTooLarge flag is always set (see
cborparser.c:174). So there's no point in checking for the flag and we
can instead go straight to the internal function.


thiagomacieira:
@erichkeane, @lpereira

lpereira:
:+1:

thiagomacieira:
Thanks!

PR #12, opened 2015-09-18
Fix/simplify printing of floating point numbers that are integer

thiagomacieira:
This avoids writing to a buffer and scanning it. As an advantage, we
always print full precision when it matches an integer.

Incidentally, let's make sure we use the C11 macro that has the correct
number of decimal digits a double-precision floating-point can hold.

### DIFF
--- a/src/cborpretty.c
+++ b/src/cborpretty.c
@@ -28,6 +28,7 @@
 
 #include <float.h>
 #include <inttypes.h>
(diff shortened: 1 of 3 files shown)


thiagomacieira:
@erichkeane , @lpereira

thiagomacieira:
Fixed.

PR #13, opened 2015-09-21
Don't tempt undefined behaviour in incrementing null pointers

thiagomacieira:
When we wrote:
       encoder->end = encoder->ptr = NULL;
    }
    ++encoder->ptr;

We're incrementing a null pointer. This could be considered undefined
behaviour and the compiler would be free to remove the increment.

This commit solves that by using an integer called bytes_needed to store
the number of bytes needed.

The trick is, however, that this commit is actually a no-op: the new
member occupies the same storage as ptr and in all of the new
conditionals, both branches evaluate to the same code, so the
conditional can be removed.

The code is complex like this to ensure we don't run afoul of another
undefined behaviour: reading from a member of an union that was not the
last written to.

CID: 306167

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -139,7 +139,10 @@ CBOR_API const char *cbor_error_string(CborError error);
 /* Encoder API */
 struct CborEncoder
 {
-    uint8_t *ptr;
+    union {
+        uint8_t *ptr;
+        ptrdiff_t bytes_needed;
+    };
     const uint8_t *end;
     int flags;
 };
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -69,15 +69,32 @@ static inline void put64(void *where, uint64_t v)
     memcpy(where, &v, sizeof(v));
 }
 
+static inline bool would_overflow(CborEncoder *encoder, size_t len)
+{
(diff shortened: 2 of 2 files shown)


thiagomacieira:
@erichkeane

tripzero:
+1

src/cbor.h
```diff
@@ -139,7 +139,10 @@ CBOR_API const char *cbor_error_string(CborError error);
 /* Encoder API */
 struct CborEncoder
 {
-    uint8_t *ptr;
+    union {
+        uint8_t *ptr;
+        ptrdiff_t bytes_needed;
+    };
```
lpereira: Are anonymous unions supported by all supported compilers?

thiagomacieira:
Hmm... C99 does not officially support anonymous unions. C11 does (6.7.2.1p13 _"An unnamed member whose type specifier is a structure specifier with no tag is called an anonymous structure; an unnamed member whose type specifier is a union specifier with no tag is called an anonymous union. The members of an anonymous structure or union are considered to be members of the containing structure or union."_)

I don't want to do casting, since that's type-punning and also results in undefined behaviour. I could make it non-anonymous, but that would break source compatibility with users of the ptr member.

Issue #4, opened 2015-07-10
Assert for writing too many items to a container, or closing a container early.


erichkeane:
When encoding, you start a container with either:
CBOR_API CborError cbor_encoder_create_array(CborEncoder *encoder, CborEncoder *arrayEncoder, size_t length);
CBOR_API CborError cbor_encoder_create_map(CborEncoder *encoder, CborEncoder *mapEncoder, size_t length);

The 'length' parameter can either be CborIndefiniteLength, or the actual value.  When that length parameter is specified, it would be nice to have an assert (either on add, or on close?) that would confirm that the correct amount of items were written.  It becomes pretty easy to forget to change the length when doing development, and it would be great to catch this rather than the otherwise unfortunate behaviors (such as failed finds in a map).

thiagomacieira:
Good idea

PR #14, opened 2015-09-21
Implement checking of array and map sizes

thiagomacieira:
This fixes issue #4. The code is placed in a new file and function so
that only people who need it will pay the cost of this function. The
impact to existing code is minimal.

Before:
   text    data     bss     dec     hex filename
    945       0       0     945     3b1 cborencoder.o
   2743       0       0    2743     ab7 cborparser.o
After:
   1060       0       0    1060     424 cborencoder.o
   2756       0       0    2756     ac4 cborparser.o

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -127,6 +127,10 @@ typedef enum CborError {
     CborErrorDuplicateObjectKeys,
     CborErrorInvalidUtf8TextString,
 
+    /* encoder errors */
+    CborErrorTooManyItems = 768,
+    CborErrorTooFewItems,
+
     /* internal implementation errors */
     CborErrorDataTooLarge = 1024,
     CborErrorNestingTooDeep,
@@ -144,6 +148,7 @@ struct CborEncoder
         ptrdiff_t bytes_needed;
     };
     const uint8_t *end;
+    size_t added;
     int flags;
 };
 typedef struct CborEncoder CborEncoder;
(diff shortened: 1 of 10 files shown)


thiagomacieira:
@erichkeane this implements a feature you requested

src/cbor.h
```diff
@@ -144,6 +148,7 @@ struct CborEncoder
         ptrdiff_t bytes_needed;
     };
     const uint8_t *end;
+    size_t added;
```
lpereira: This breaks ABI, is it OK? (Not sure about TinyCBOR's ABI compatibility promises.)

src/compilersupport_p.h
```diff
@@ -33,6 +33,14 @@
 #include <stddef.h>
 #include <stdint.h>
 
+#if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_assert >= 200410
+#  define cbor_static_assert(x)         static_assert(x, #x)
+#elif !defined(__cplusplus) && defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 406)
+#  define cbor_static_assert(x)         _Static_assert(x, #x)
+#else
+#  define cbor_static_assert(x)         ((void)sizeof(struct { int m : 2*!!(x) - 1; }))
```
lpereira: Nice. I usually see this defined in terms of `sizeof()` of an array with negative number of elements.

lpereira:
Looks OK from a cursory glance (at first I thought that only numbers and containers were counted, but other types are counted indirectly).

I like the feature; will ensure my code use it as soon as this gets merged.

src/cbor.h
thiagomacieira: Yes, BC break is fine, we've only got a static library anyway. SC is more important.

src/compilersupport_p.h
thiagomacieira: I tried that, but neither Clang nor ICC complain about an array of size -1. 

Original code:
```c
#  define cbor_static_assert(x, msg)        (void)(sizeof(char[(!!(msg) * !!(x)) ? 1 : -1]))
```

If you remove that (void) and return cbor_static_assert(0, "Hello") from function, ICC:

```asm
        movl      $-1, %eax
        ret       
```

Clang:
```asm
        movl    $4294967295, %eax       # imm = 0xFFFFFFFF
        popq    %rbp
        retq
```

So I copied the solution from the kernel (BUILD_BUG_ON).

PR #15, opened 2015-09-25
Build fix when using a C compiler

lpereira:
Check if __cplusplus is defined before checking its value.

Signed-off-by: Leandro Pereira <leandro.pereira@intel.com>

### DIFF
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -33,7 +33,7 @@
 #include <stddef.h>
 #include <stdint.h>
 
-#if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_assert >= 200410
+#if __STDC_VERSION__ >= 201112L || (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(__cpp_static_assert) && __cpp_static_assert >= 200410)
 #  define cbor_static_assert(x)         static_assert(x, #x)
 #elif !defined(__cplusplus) && defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 406)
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Patch is good but unnecessary.

```C99 section 6.10.1 "Conditional inclusion" paragraph 3
Prior to evaluation, macro invocations in the list of preprocessing tokens that will become the controlling constant expression are replaced (...), just as in normal text. [...] After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers are replaced with the pp-number 0, and then each preprocessing token is converted into a token.
```

So an undefined __cplusplus will expand to 0. Since the check is for larger than a non-zero quantity, this should not break the build.

lpereira:
[**@thiagomacieira**](https://github.com/thiagomacieira) commented on [Sep 25, 2015, 12:44 PM GMT-3](https://github.com/01org/tinycbor/pull/15#issuecomment-143257348 "2015-09-25T15:44:33Z - Replied by Github Comment Enhancer"):

> 
> So an undefined __cplusplus will expand to 0. Since the check is for larger than a non-zero quantity, this should not break the build.

Tell that to GCC 5.1.1. :)

thiagomacieira:
No problems here with GCC 5.2.1. What exactly is wrong?

lpereira:
I'm closing this PR.

I'm building TinyCBOR as part of Soletta which uses `-Wundef` by default -- and I always develop with `-Werror` on:

    error: "__cpp_static_assert" is not defined [-Wundef]
     #if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_asser

I'm adding `-Wno-undef` for TinyCBOR in the Soletta build as this is harmless; the exception will be removed as soon as TinyCBOR's `Makefile` is used.

PR #18, opened 2015-10-29
Fix a parsing issue caused by pointer wraparounds

thiagomacieira:
If the size of the string was a suitable value, the "ptr + total" check
could overflow and we wouldn't know because the sum would most likely be
smaller than value->parser->end. We would then proceed to allocate a
silly amount of memory.

I never caught this in my own testing because, on Linux, the heap
usually starts in the low 4 GB of the address space, so adding
0xffffffff000000000 would not cause an overflow. However, on OS X, the
heap starts above 4 GB, so the pointer does wrap around.

So test specifically for the pointer wraparound cases, plus the case of
sizes larger than the address space in 32-bit systems.

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -108,7 +108,7 @@ static CborError preparse_value(CborValue *it)
     }
 
     size_t bytesNeeded = descriptor < Value8Bit ? 0 : (1 << (descriptor - Value8Bit));
-    if (it->ptr + 1 + bytesNeeded > parser->end)
+    if (bytesNeeded + 1 > (size_t)(parser->end - it->ptr))
         return CborErrorUnexpectedEOF;
 
     uint8_t majortype = type >> MajorTypeShift;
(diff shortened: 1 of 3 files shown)


thiagomacieira:
@erichkeane, @lpereira

src/cborparser.c
```diff
@@ -108,7 +108,7 @@ static CborError preparse_value(CborValue *it)
     }
 
     size_t bytesNeeded = descriptor < Value8Bit ? 0 : (1 << (descriptor - Value8Bit));
```
lpereira: Is the `1` being left-shifted going to be interpreted as an `int` or as a `size_t`?

src/cborparser.c
```diff
@@ -521,7 +521,7 @@ static CborError iterate_string_chunks(const CborValue *value, char *buffer, siz
         err = extract_length(value->parser, &ptr, &total);
         if (err)
             return err;
-        if (ptr + total > value->parser->end)
+        if (total > (size_t)(value->parser->end - ptr))
```
lpereira: Maybe we're into a `untrusted_value_greater_than()` primitive?

src/cborparser.c
thiagomacieira: 1 is int, but the value of descriptor is already range-checked. By construction, it's in the range 0-27. So bytesNeeded is either 0, 1, 2, 4, or 8. That's an int.

That int is extended to size_t, but there's no sign conversion.

src/cborparser.c
thiagomacieira: What would it do?

PR #22, opened 2016-02-24
parser: Set the length to 0 on error in extract_length()

### LINKED ISSUE #21: cborparser.c: Compilation warning with ARM GCC and -Os
I'm using tinycbor in a ARM project that uses `-Wall -Werror -Os`. Here is a minimal example that shows the warning that I'm getting: 

```
 $ arm-none-eabi-gcc -Wall -Werror -Os -c cborparser.c
cborparser.c: In function 'iterate_string_chunks':
cborparser.c:563:27: error: 'chunkLen' may be used uninitialized in this function [-Werror=maybe-uninitialized]
                 *result = func(buffer + total, ptr, chunkLen);
                           ^
cborparser.c:527:23: error: 'total' may be used uninitialized in this function [-Werror=maybe-uninitialized]
             *result = func(buffer, ptr, total);
                       ^
cc1: all warnings being treated as errors
```

I understand this is totally GCC's fault. [size_t total](https://github.com/01org/tinycbor/blob/master/src/cborparser.c#L516) and [size_t chunkLen](https://github.com/01org/tinycbor/blob/master/src/cborparser.c#L537) are not used if `extract_length()` returns an error. I could suppress the warning, but I would like to just use tinycbor in the rest of my project without making a special case for it. 

I considered just initializing the variables, but in this case I believe it's better to just set the length to 0 in `extract_length()` when there is an error, which shouldn't happen too often and solves my problem. I'll send a patch request in a moment that does that.

colagrosso:
Fixes #21

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -65,8 +65,10 @@ static CborError extract_length(const CborParser *parser, const uint8_t **ptr, s
 {
     uint64_t v;
     CborError err = extract_number(ptr, parser->end, &v);
-    if (err)
+    if (err) {
+        *len = 0;
         return err;
+    }
 
     *len = v;
     if (v != *len)


thiagomacieira:
Looks good, but would you write a little more in the commit body explaining why you're doing it? Just so we don't have to look up the bug report to know why you've done it.

Something like "This silences a false-positive warning in iterate_string_chunks() that some variables were used without initialisation".

colagrosso:
Thanks for the quick review.

> Looks good, but would you write a little more in the commit body explaining why you're doing it?

Good call. Done.

thiagomacieira:
Sorry, I forgot. Can you add a Signed-off-by to the commit with your name?

colagrosso:
> Can you add a Signed-off-by to the commit with your name?

No prob. I see that line now in the repo's other commits.

colagrosso:
Thanks for the suggestions and quick reviews. I'm closing the PR because the change has been committed with your help at 629d5b7.

PR #24, opened 2016-04-12
Fix the simplereader example after the API change a while ago

### LINKED ISSUE #23: Trying to  compile example: a lot of errors, even with C99 compiler
examples/simplereader.c:74:19: warning: implicit declaration of function 'cbor_value_dup_string' is invalid in C99 [-Wimplicit-function-declaration]
            err = cbor_value_dup_string(it, &buf, &n, it);
                  ^
examples/simplereader.c:164:38: warning: passing 'char *' to parameter of type 'const uint8_t *' (aka 'const unsigned char *') converts between pointers to integer types with different sign [-Wpointer-sign]
    CborError err = cbor_parser_init(buf, length, 0, &parser, &it);
                                     ^~~
examples/../src/cbor.h:223:52: note: passing argument to parameter 'buffer' here
CBOR_API CborError cbor_parser_init(const uint8_t *buffer, size_t size, int flags, CborParser *parser, CborValue *it);
                                                   ^
examples/simplereader.c:171:24: error: 'const uint8_t *' (aka 'const unsigned char *') and 'char *' are not pointers to compatible types
                it.ptr - buf, cbor_error_string(err));
                ~~~~~~ ^ ~~~
2 warnings and 1 error generated.

Also, it's not very Visual Studio friendly, if at all. I've been able to compile it under studio with some changes, but then your example is unable to dump data from CBOR.

It starts print "Map[" and then blows with unexpected EOF error.

thiagomacieira:
We changed a lot of "char*" that was binary to "uint8_t*". The example
was never fixed.

This fixes #23.

### DIFF
--- a/examples/simplereader.c
+++ b/examples/simplereader.c
@@ -6,7 +6,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 
-static char *readfile(const char *fname, size_t *size)
+static uint8_t *readfile(const char *fname, size_t *size)
 {
     struct stat st;
     FILE *f = fopen(fname, "rb");
@@ -14,7 +14,7 @@ static char *readfile(const char *fname, size_t *size)
         return NULL;
     if (fstat(fileno(f), &st) == -1)
         return NULL;
-    char *buf = malloc(st.st_size);
+    uint8_t *buf = malloc(st.st_size);
     *size = fread(buf, st.st_size, 1, f);
     fclose(f);
     return buf;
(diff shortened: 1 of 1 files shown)


thiagomacieira:
@lpereira, @bialix, please take a look

lpereira:
LGTM. I'd just wait @bialix confirm it works with MSVC before merging.

thiagomacieira:
I'm not waiting for MSVC because this is not a bugfix to whatever issues there were with that compiler: the bug report didn't give enough details for me to take any action.

bialix:
Sorry, I'm not interested in this library anymore, and it's not so simple to build it on Windows, so I retreat.

thiagomacieira:
@bialix, ok, that's your choice. But please note that we can't help if we don't know what the problem is and you never explained what your Windows problems were.

PR #25, opened 2016-06-03
Fixes for Visual Studio

dthaler:
Signed-off-by: Dave Thaler <dthaler@microsoft.com>

### DIFF
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -41,7 +41,7 @@ void cbor_encoder_init(CborEncoder *encoder, uint8_t *buffer, size_t size, int f
     encoder->flags = flags;
 }
 
-static inline void put16(void *where, uint16_t v)
+CBOR_INLINE_API void put16(void *where, uint16_t v)
 {
     v = cbor_htons(v);
     memcpy(where, &v, sizeof(v));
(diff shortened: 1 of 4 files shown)


src/cborencoder.c
```diff
@@ -41,7 +41,7 @@ void cbor_encoder_init(CborEncoder *encoder, uint8_t *buffer, size_t size, int f
     encoder->flags = flags;
 }
 
-static inline void put16(void *where, uint16_t v)
+CBOR_INLINE_API void put16(void *where, uint16_t v)
```
thiagomacieira: This function isn't API, so I'd rather not have it using the "CBOR_*API" macros.

What was the issue? My guess: MSVC 2012 in C mode does not understand "inline". Correct?

thiagomacieira:
Sorry, I don't like the change. If the problem is the inline keyword, we can #define it in compilersupport_p.h (_p means "private header" and is not API)

thiagomacieira:
Can you try these two commits instead:
* https://github.com/thiagomacieira/tinycbor/commit/6523afd8e26c5e980580f8f92d24a8854afe2cfa
* https://github.com/thiagomacieira/tinycbor/commit/92a25333a4344c1bb7638f03c0262764e8b1779a

Note I've updated the master branch with fixes for -W3 warnings with MSVC 2015.

thiagomacieira:
Probably replaced by #26

thiagomacieira:
Replaced by #27

PR #26, opened 2016-06-07
Fix msvc2012

thiagomacieira:
These commits attempt at fixing the build with MSVC 2012.

@dthaler


thiagomacieira:
Tested with VS 2013 & 2015: compiles fine.

2012 complains about lack of stdbool.h

thiagomacieira:
@dantler

PR #27, opened 2016-06-08
Fix build msvc

thiagomacieira:
This series of commits fixes the build and makes the parser and encoder tests pass with MSVC.

The to and from JSON functionality is not enabled with MSVC.

@dantler, @dthaler


dthaler:
Verified on Windows with VS2013.  Looks good to me.

PR #28, opened 2016-07-07
Fix compilation for systems without math.h header

otaviobp:
cbortojson and cborpretty depends on decode_half function from
compilersupport_p.h that uses math.h functions. As cbortojson and
cborpretty are usually important only in development phase, move math
dependency to another header so it is now possible to compile tinycbor
in systems that doesn't provide math.h header

Signed-off-by: Otavio Pontes <otavio.pontes@intel.com>

### DIFF
--- a/src/cborpretty.c
+++ b/src/cborpretty.c
@@ -25,6 +25,7 @@
 #define _BSD_SOURCE 1
 #include "cbor.h"
 #include "compilersupport_p.h"
+#include "math_support_p.h"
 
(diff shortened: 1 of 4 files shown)


thiagomacieira:
Looks good, but please name the header math_support_p.h (the _p indicating it's private API, not meant for users).

otaviobp:
Header name changed.

thiagomacieira:
Accepted as 358a7e4

Issue #32, opened 2016-07-21
Compilation fails due to the usage of a file descriptor above 9 for redirection


ffontaine:
In Makefile, you're using the file descriptor 10, to redirect input to .config:
$(MAKE) -f $(SRCDIR)Makefile.configure OUT='>&10' configure 10> $@

This fails on my PC. I advise you to use a file descriptor below 10 like 9 as "Redirections using file descriptors greater than 9 should be used with care, as they may conflict with file descriptors the shell uses internally." (cf. https://www.gnu.org/software/bash/manual/html_node/Redirections.html)

PR #33, opened 2016-07-21
Fix compilation error due to file descriptor

ffontaine:
In Makefile, the file descriptor 10 is used to redirect input to .config:
$(MAKE) -f $(SRCDIR)Makefile.configure OUT='>&10' configure 10> $@

This patch replaces this file descriptor from to 10 to 9 as "Redirections using file descriptors greater than 9 should be used with care, as they may conflict with file descriptors the shell uses internally." (cf. https://www.gnu.org/software/bash/manual/html_node/Redirections.html)


thiagomacieira:
Merged as f775cf5. Fixes #32

PR #35, opened 2016-07-30
Give a name to unnamed union in CborEncoder

ffontaine:
cbor.h contains an unnamed union in CborEncoder structure, this functionality is
not always supported by all compilers, especially it seems that the Blackfin
build on buildroot fails (with gcc 4.3). This patch names this union "data".

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -162,7 +162,7 @@ struct CborEncoder
     union {
         uint8_t *ptr;
         ptrdiff_t bytes_needed;
-    };
+    } data;
     const uint8_t *end;
     size_t added;
     int flags;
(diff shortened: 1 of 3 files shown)


thiagomacieira:
Interesting, I thought anonymous unions were a C99 feature, but they're actually C11.

thiagomacieira:
Merged as 0e1440a

thiagomacieira:
Reverting this in the master branch. I'm re-applying to the dev branch.

thiagomacieira:
Re-applied with fixes to dev branch, as ede7f1431ae06c9086f2a83a57bd7832d99280e3.

Issue #42, opened 2016-08-16
fopencookie is not always available


ffontaine:
Currently, tinycbor checks in Makefile.configure the availability of open_memstream in the toolchain to compile src/open_memstream.c. However, this file is also using fopencookie which is not always available as fopencookie is not a part of any standard (it's a GNU function and open_memstream is POSIX). If __GLIBC__ is defined, the availability of this function should be checked.

Moreover, funopen is tested in Makefile.configure but never used in Makefile. Indeed, funopen is called depending on __APPLE__ being defined, perhaps this test should be removed from Makefile.configure?

thiagomacieira:
Just describe the problem you're having.

ffontaine:
Tinycbor 0.3.2 with patch on unnamed union is failing on buildroot with bfin target. More info can be found here: https://patchwork.ozlabs.org/patch/659398/.

ffontaine:
Here is more information: bfin target uses uclinux which has been configured without fopencookie, here is the stdio.h from uclinux:
```

#ifdef  __USE_GNU
#ifdef __UCLIBC_HAS_GLIBC_CUSTOM_STREAMS__
/* Create a new stream that refers to the given magic cookie,
   and uses the given functions for input and output.  */
extern FILE *fopencookie (void *__restrict __magic_cookie,
                          __const char *__restrict __modes,
                          _IO_cookie_io_functions_t __io_funcs) __THROW __wur;

/* Create a new stream that refers to a memory buffer.  */
extern FILE *fmemopen (void *__s, size_t __len, __const char *__modes)
  __THROW __wur;

/* Open a stream that writes into a malloc'd buffer that is expanded as
   necessary.  *BUFLOC and *SIZELOC are updated with the buffer's location
   and the number of characters written on fflush or fclose.  */
extern FILE *open_memstream (char **__restrict __bufloc,
                             size_t *__restrict __sizeloc) __THROW __wur;
#endif
#endif
```

I don't think it's possible to implement a version of open_memstream without fopencookie or funopen. So, I think we'll have to disable tinycbor for bfin target. You can close this issue.

thiagomacieira:
Not really. You're right that the check was wrong. We checked for A in the Makefile, then for B in the source code. That should be fixed.

Also, you don't need to disable the entire library. You will only need to disable the part about converting to JSON. Though I can also accept a patch that degrades that functionality and returns an internal error or "not implemented" for when it would have been necessary.

PR #43, opened 2016-08-30
Properly manage open_memstream availability

ffontaine:
- Update Makefile.configure to test the availability of fopencookie
- If open_memstream is not available on the system AND funopen or fopencookie are available, compile our own version of open_memstream
- If open_memstream can not be compiled, return a new error (CborErrorJsonNotImplemented)

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -149,6 +149,7 @@ typedef enum CborError {
     /* errors in converting to JSON */
     CborErrorJsonObjectKeyIsAggregate,
     CborErrorJsonObjectKeyNotString,
+    CborErrorJsonNotImplemented,
 
     CborErrorOutOfMemory = ~0U / 2 + 1,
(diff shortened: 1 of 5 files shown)


src/cbor.h
```diff
@@ -147,6 +147,7 @@ typedef enum CborError {
     CborErrorUnsupportedType,
 
     /* errors in converting to JSON */
+    CborErrorJsonNotImplemented,
```
thiagomacieira: Please add it after the other Json entries, so as to not change the values of the existing enums.

thiagomacieira:
Looks good otherwise. Can you just rearrange as I noted?

thiagomacieira:
Oh, please write the words "fixes" and "#42" together in the commit message.

ffontaine:
OK, done.

thiagomacieira:
Committed as 04ada58

Issue #45, opened 2016-10-06
C4244 warnings when building in Visual Studio


alradmsft:
Hello,

I'm building the lib in Visual Studio with /W4 switch. I'm getting the warnings below. I understand that in your code you do have checks for data loss during those conversions; however, I have to have code building clean in my environment. I can put together a change to disable the warning in those specific places. Please let me know if that's OK.  

X64:
cborparser.c(802): warning C4244: '=': conversion from 'uint64_t' to 'int', possible loss of data
cborparser.c(808): warning C4244: '=': conversion from 'uint64_t' to 'int', possible loss of data

X86:

cbor.h(166): warning C4201: nonstandard extension used: nameless struct/union
cbor.h(364): warning C4244: '=': conversion from 'uint64_t' to 'std::size_t', possible loss of data
cbor.h(419): warning C4244: '=': conversion from 'uint64_t' to 'std::size_t', possible loss of data
cbor.h(431): warning C4244: '=': conversion from 'uint64_t' to 'std::size_t', possible loss of data

thiagomacieira:
The non-standard extension is fixed by 4628a110b736c3c31bd8c577ea1d29400e0c0a53 (which got into the dev branch as ede7f1431ae06c9086f2a83a57bd7832d99280e3). The other issues are easily solved with an explicit cast.

thiagomacieira:
Can you check if 806f0c9 (from https://github.com/thiagomacieira/tinycbor) solves the issue?

alradmsft:
Yes, that change fixes the the warnings above. Thanks! However, I didn't paste one more from last time (sorry).

compilersupport_p.h(213): warning C4244: 'return': conversion from 'int' to 'unsigned short', possible loss of data

thiagomacieira:
Were there no warnings on lines 190-192 of compilersupport_p.h? That assigns from an uint64_t to int, though the integer operations on the RHS clearly limit the width.

alradmsft:
Hmm... they don't. It looks like only single bit shifts trigger the warning, but not when you have two of them chained. I've reduced the problem to the code below. I will check with our compiler team as to why there is a difference. Will get back to you. 

#include <stdio.h>
#include <stdint.h>

int main()
{
	uint64_t v = 123;
	int i = v;
	int j = v >> 1;
	int k = v << 1;
	int m = v >> 63 << 15;
	wprintf(L"i = %d, j = %d, k = %d, m = %d\n", i, j, k, m);
	return 0;
}

[...]

PR #48, opened 2016-10-12
Stop-gap fixes for compiler warnings

alradmsft:
Related to #45. This is just to show the lines that I had to change to get a clean build on VS2015 compiler.

### DIFF
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -290,7 +290,7 @@ static inline CborError encode_number_no_update(CborEncoder *encoder, uint64_t u
     if (ui < Value8Bit) {
         *bufstart += shiftedMajorType;
     } else {
-        unsigned more = 0;
+		uint8_t more = 0;
         if (ui > 0xffU)
             ++more;
         if (ui > 0xffffU)
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -210,7 +210,15 @@ static inline unsigned short encode_half(double val)
         /* underflow, make zero */
         return 0;
     }
+
+//
+// bugbug: working with tinycbor folks to address this properly. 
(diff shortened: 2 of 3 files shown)


src/cborencoder.c
```diff
@@ -290,7 +290,7 @@ static inline CborError encode_number_no_update(CborEncoder *encoder, uint64_t u
     if (ui < Value8Bit) {
         *bufstart += shiftedMajorType;
     } else {
-        unsigned more = 0;
+		uint8_t more = 0;
```
thiagomacieira: Please fix the whitespace here (TinyCBOR uses only spaces, no tab).

src/compilersupport_p.h
```diff
@@ -210,7 +210,15 @@ static inline unsigned short encode_half(double val)
         /* underflow, make zero */
         return 0;
     }
+
+//
+// bugbug: working with tinycbor folks to address this properly. 
+// Issue https://github.com/01org/tinycbor/issues/47
+//
+#pragma warning(disable:4244)
```
thiagomacieira: I can't accept an MSVC pragma here, not without #ifndef _MSC_VER. Is there no other way of getting this? Maybe

  return (unsigned short)(sign | ((exp + 15) << 10) | mant);

alradmsft:
Sorry I didn't realize that you were going to pull this actually. I thought you only were looking for the list of warnings so I hacked something up really quick. You got me embarrassed with spaces actually. :-)

I can work on the proper change, no problem. For the compilersupport_p.h, I used pragma to just get my build working so I can play with the lib. Doing a cast + comment as to why it is safe is OK. I did look at DirectX implementation of this function and uses all unsigned integers so it doesn't have this problem. But I personally don't care for such a major re-write.

alradmsft:
Will start another PR.

PR #53, opened 2016-11-02
Fixed duplicate definitions for targets with certain macros already d…

jprestwo:
…efined

Signed-off-by: James Prestwood <james.prestwood@intel.com>

### DIFF
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -61,7 +61,9 @@
 #  define inline    CBOR_INLINE
 #endif
 
+#ifndef STRINGIFY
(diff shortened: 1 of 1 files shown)


jprestwo:
Zephyr already has these macro's defined, and with its strict compile options the warnings turn into errors. Wrapping these definitions in ifdef's solves the problem.

poussa:
+1

thiagomacieira:
Merged as eed8aa4. Note that it's poor practice for Zephyr system headers to leave such unnamespaced macros defined...

PR #57, opened 2016-11-10
Create a Perl script that properly creates tags

thiagomacieira:
The previous part of the Makefile didn't work properly. Instead of using
the contents of the VERSION file, it updated the file right before the
tagging and always for the next minor release.

Instead, move everything to a Perl script that updates the VERSION file
only after the tagging and sets up for a patch release (which are more
frequent). Updating of the dev branch is left as an exercise for the
reader.

As a bonus, this will work on Windows too without Unix shell tools.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,29 +177,8 @@ distcheck: .git
 	cd $${TMPDIR-/tmp}/tinycbor-distcheck && $(MAKE) silentcheck
 	$(RM) -r $${TMPDIR-/tmp}/tinycbor-distcheck
 
-release: .git
-	$(MAKE) -f $(MAKEFILE) distcheck
-ifeq ($(VERSION),)
-	git -C $(SRCDIR). show HEAD:VERSION | \
-	  perl -l -n -e '@_ = split /\./; print "$$_[0]." . ($$_[1] + 1)' > $(SRCDIR)VERSION
-else
-	echo "$(VERSION)" > VERSION
-endif
-	if test -n "`git diff VERSION`"; then \
-	  git -C $(SRCDIR). commit -s -m "Update version number" VERSION; \
-	fi
-	{ echo "TinyCBOR release `cat $(SRCDIR)VERSION`"; \
-	  echo; \
-	  echo '# Write something nice about this release here'; \
-	  tmpl=`git -C $(SRCDIR). config --get commit.template` && \
-		cat "$$tmpl"; \
-	  echo '# Commit log:'; \
-	  git -C $(SRCDIR). shortlog -e --no-merges HEAD --not `git -C $(SRCDIR). tag` | sed 's,^,#   ,'; \
-	  echo '# Header diff:'; \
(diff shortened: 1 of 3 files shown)


thiagomacieira:
@dantler, your opinion please.

dantler:
Windows doesn't have perl installed by default.  I got this error in ActiveState Perl:

```
Experimental push on scalar is now forbidden at maketag.pl line 60, near "'1';"
Execution of maketag.pl aborted due to compilation errors.
make: *** [tag] Error 255
```

As an FYI, the Makefile needs a lot of changes to be compatible with Visual Studio.  To test this, I removed `distcheck` as a precursor to `tag` in the Makefile.

thiagomacieira:
> Windows doesn't have perl installed by default. I got this error in ActiveState Perl:
> ```
> Experimental push on scalar is now forbidden at maketag.pl line 60, near "'1';"
> Execution of maketag.pl aborted due to compilation errors.
> make: *** [tag] Error 255
> ```

Fixed.

> As an FYI, the Makefile needs a lot of changes to be compatible with Visual Studio.

Please use Makefile.nmake instead.

dantler:
Thanks for the nmake tip.  I didn't see that before.  Now I'm using this:  `nmake tag /f Makefile.nmake`

However I hit a new issue.  Someone is calling your "run" function on "vi".

```
Cannot run vi: No such file or directory at maketag.pl line 4.
NMAKE : fatal error U1077: 'C:\Perl64\bin\perl.EXE' : return code '0x2'
Stop.
```

I figured out that it was my GIT_EDITOR causing this problem.  If I change it to 'vim', which I do have installed, the script hangs after printing this message: `Vim: Warning: Output is not to a terminal`.  This is an annoying problem to solve!  I guess the way perl is launching vim doesn't cooperate with conhost.exe.

If I set GIT_EDITOR to GVIM, I get a window to allow me to edit.  Looks good to me!

```
set GIT_EDITOR=gvim
```

Maybe GIT_EDITOR can default to notepad.exe on Windows if there is no other option specified?

thiagomacieira:
> Thanks for the nmake tip. I didn't see that before. Now I'm using this: nmake tag /f Makefile.nmake
>
> However I hit a new issue. Someone is calling your "run" function on "vi".

That's not a new issue then because I didn't change how it finds your editor. The code asks Git what it thinks your editor is. The defaults come from Git itself and can be overridden by several config files (config entry "core.editor") and by the environment variable GIT_EDITOR.

> If I change it to 'vim', which I do have installed, the script hangs after printing this message: Vim: Warning: Output is not to a terminal. This is an annoying problem to solve! I guess the way perl is launching vim doesn't cooperate with conhost.exe.

Right, it's opening via pipe, which is probably the problem. I can probably change it to run without capturing output and just forwarding stdout instead. Fixing that.

> Maybe GIT_EDITOR can default to notepad.exe on Windows if there is no other option specified?

That's up to Git itself.

dantler:
Your pipe fix worked, @thiagomacieira .  Thanks!  

👍

PR #59, opened 2016-11-30
Buildsystem update

thiagomacieira:
Just a few quick fixes.

@erichkeane, @dantler, @lpereira


erichkeane:
Don't totally understand the build stuff, but this seems fine. +1.

PR #60, opened 2016-11-30
Add a way to replace the call to malloc with a different allocator

thiagomacieira:
Just define the macro CBOR_ALLOC to your allocator when compiling the
source file cbor_parser_dup_string.c. The function will call that
allocator instead of malloc.

You probably also want to define CBOR_ALLOC_INCLUDE to the header file
that contains the declaration of that function.

### DIFF
--- a/src/cborparser_dup_string.c
+++ b/src/cborparser_dup_string.c
@@ -29,7 +29,15 @@
 #endif
 
 #include "cbor.h"
-#include <stdlib.h>
+#ifdef CBOR_ALLOC_INCLUDE
+#  include CBOR_ALLOC_INCLUDE
+#else
+#  include <stdlib.h>
+#endif
+
+#ifndef CBOR_ALLOC
+#  define CBOR_ALLOC        malloc
+#endif
 
 /**
  * \fn CborError cbor_value_dup_text_string(const CborValue *value, char **buffer, size_t *buflen, CborValue *next)
@@ -99,7 +107,7 @@ CborError _cbor_value_dup_string(const CborValue *value, void **buffer, size_t *
         return err;
 
     ++*buflen;
-    *buffer = malloc(*buflen);
+    *buffer = CBOR_ALLOC(*buflen);
     if (!*buffer) {
         /* out of memory */
         return CborErrorOutOfMemory;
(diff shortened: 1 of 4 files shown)


thiagomacieira:
@lpereira, @dantler

lpereira:
I'm having a hard time seeing a reason for something like this. Why would this be desired on a general-purpose library? Today you're only supporting mallloc(), and maybe tomorrow you'll need realloc(), calloc(), and other memory-allocation functions...

If `malloc()` isn't available on a platform, have the user pass a pre-allocated buffer that has been allocated using whatever means they have available, and return an error if it's not large enough. Marking such function with the `warn_unused_result` attribute (if building with GCC) will help catch uses where the result value was ignored.

erichkeane:
> I'm having a hard time seeing a reason for something like this. Why would this be desired on a general-purpose library? Today you're only supporting mallloc(), and maybe tomorrow you'll need realloc(), calloc(), and other memory-allocation functions...

I agree.  This seems like a tougher to use and likely incomplete solution.  I think that a better solution here might be to have functions that NEED this have c-sytle-overloads (ie, with different names;)) take the allocator as a FP.

>If malloc() isn't available on a platform, have the user pass a pre-allocated buffer that has been allocated using whatever means they have available, and return an error if it's not large enough. Marking such function with the warn_unused_result attribute (if building with GCC) will help catch uses where the result value was ignored.

Isn't that our behavior in the non-string dupe case?  We already have that functionality in the copy-string, so I don't think that is all that necessary.

ossama-othman:
@lpereira It's not a matter of whether or not `malloc()` is available on the platform.  Rather, there are times when it is more desirable to use a different allocator.  Regardless, I agree that having the user pass in a pre-allocated buffer is preferable to avoid potential allocator mismatches.

@thiagomacieira Does the `cbor_value_dup_*_string()` family of functions really buy the user anything?  Why not just have the user handling memory allocation for the string to be dup-ed, and make them use `cbor_value_copy_text_string()` or `cbor_value_copy_byte_string()`, accordingly?

thiagomacieira:
@ossama-othman: yeah, I agree that they should just use the "copy" functions instead. The source code for the "dup" function is qutie simple (and dumb). Might as well have IoTivity use them instead of the "dup" functions.

thiagomacieira:
Rejecting my own pull.

PR #61, opened 2017-01-18
Support C99 standards (enums, assert)

rzr:
Cast enum values to int to prevent errors like:
   deps/tinycbor/src/cborconstants_p.h:47:31:\
   error: ISO C restricts enumerator values to range of ‘int’\
   [-Werror=pedantic]
   MajorTypeMask           = ~0U << MajorTypeShift,

"error: ISO C99 does not support" ‘Static_assert’

This effort is needed for RIOT port of iotivity-contrained

Change-Id: Ib8addf7c6703298faecfc4006cf2a9e3c1036d4c
Bug-RIOT: https://github.com/RIOT-OS/RIOT/issues/6241

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -151,8 +151,8 @@ typedef enum CborError {
     CborErrorJsonObjectKeyNotString,
     CborErrorJsonNotImplemented,
 
-    CborErrorOutOfMemory = ~0U / 2 + 1,
-    CborErrorInternalError = ~0U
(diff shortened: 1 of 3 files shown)


rzr:
cc: @kmaloor @thiagomacieira

thiagomacieira:
Thank you, the change looks good. But I need a Signed-off-by line. Can you sign your commit, please?

Drop the Change-Id line while you're at it, this is not JIRA.

thiagomacieira:
Merged as 3e23929

PR #62, opened 2017-03-04
Fix stack corruption due to calling convention mismatch

alradmsft:
Standard library implementations may specify calling convention for memcpy()
explicitly. However, the cbor* APIs don't do that. If you compile the lib
with default calling convention that doesn't match the calling convention of
memcpy(), the iterate_string_chunks() will not setup the stack for memcpy()
call correctly resulting in a stack corruption. The compiler doesn't catch
this issue because of the cast that is being applied when passing memcpy()
to the iterate_string_chunks().

The fix is to wrap mempy() in a function that conforms to the declaration of
the IterateFunction which does two things, removes the need for a cast and
uncouples memcpy() calling convention from the rest of the cbor* APIs.

Signed-off-by: Alex Radutskiy (alex.radutskiy@microsoft.com)

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -915,6 +915,12 @@ static uintptr_t iterate_memcmp(char *s1, const uint8_t *s2, size_t len)
     return memcmp(s1, (const char *)s2, len) == 0;
 }
 
+static uintptr_t iterate_memcpy(char *dest, const uint8_t *src, size_t len)
+{
+    memcpy(dest, src, len);
+    return true;
+}
+
 static CborError iterate_string_chunks(const CborValue *value, char *buffer, size_t *buflen,
                                        bool *result, CborValue *next, IterateFunction func)
 {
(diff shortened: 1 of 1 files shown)


src/cborparser.c
```diff
@@ -915,6 +915,12 @@ static uintptr_t iterate_memcmp(char *s1, const uint8_t *s2, size_t len)
     return memcmp(s1, (const char *)s2, len) == 0;
 }
 
+static uintptr_t iterate_memcpy(char *dest, const uint8_t *src, size_t len)
```
erichkeane: Why does this return true, and not the actual return value of memcpy?  This conversion of bool to int could very likely prevent at least 1 major compiler from eliding this wrapper.

thiagomacieira: Yes, please return the cast return. This will still generate a function with a jmp, which is something I was trying to avoid. I wish we could detect the calling convention...

alradmsft: Ah sorry, I was staring at the bool check in iterate_string_chunks() when decided to just always return true to make return unconditional.

thiagomacieira:
Next time, please squash the two commits together when updating with requested changes. I'll do it for you now.

thiagomacieira:
Merged as acf202a38118ed049785d210b5de84effa43c3de

alradmsft:
Sorry. Will squash next time. Still getting acquainted with rules/etiquette for git/GitHub. Appreciate you guys being patient with me here.

PR #63, opened 2017-03-05
Don't cast an actual string literal to uint8_t*

thiagomacieira:
This removes an ugly cast from the source and an unnecessary 1-byte
variable from the .rodata section.

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -977,8 +977,10 @@ static CborError iterate_string_chunks(const CborValue *value, char *buffer, siz
     }
 
     /* is there enough room for the ending NUL byte? */
-    if (*result && *buflen > total)
-        *result = !!func(buffer + total, (const uint8_t *)"", 1);
+    if (*result && *buflen > total) {
+        uint8_t nul[] = { 0 };
+        *result = !!func(buffer + total, nul, 1);
+    }
     *buflen = total;
 
     if (next) {


thiagomacieira:
@erichkeane @lpereira comments?

lpereira:
LGTM.

erichkeane:
There is an obvious runtime cost to this, and if the target function keeps a pointer to this string, it will now be UB if it accesses after this block.

If the 2nd concern isn't an issue, and the rodata to stack cost is acceptable, Ive got no issue.

lpereira:
Indeed, I haven't thought of the client-using-the-ptr case. That will be an issue. (I've only used the iterators to consume the values immediately.)

thiagomacieira:
The pointer cannot leave the internal TinyCBOR code: the called function is not user code, it's either a noop, memcpy or memcmp. 

The cost of writing a single byte to the stack and getting its address is essentially the same as loading a full pointer to a register.

erichkeane:
Alright, sounds good. Just making sure it was consideredOn Mar 5, 2017 8:30 AM, Thiago Macieira <notifications@github.com> wrote:The pointer cannot leave the internal TinyCBOR code: the called function is not user code, it's either a noop, memcpy or memcmp.
The cost of writing a single byte to the stack and getting its address is essentially the same as loading a full pointer to a register.

PR #65, opened 2017-03-06
Chunked string parsing

thiagomacieira:
This is a new v0.5 feature: the ability to parse chunked CBOR strings per chunk. It also adds the ability to do zero-copy parsing of strings. From the docs:

```c
   if (cbor_value_is_text_string(value)) {
       char *ptr;
       size_t len;
       while (1) {
           err = cbor_value_get_text_string(value, &ptr, &len, &value));
           if (err) return err;
           if (ptr == NULL) return CborNoError;
           consume(ptr, len);
       }
   }
```

There are two commits in this series:

1. Add support for parsing chunked strings
2. rewrite the iterate_string_chunks function with new chunk code


thiagomacieira:
@lpereira @otaviobp @rzr @erichkeane @alradmsft

thiagomacieira:
Wrong target branch. This is meant for dev.

PR #66, opened 2017-03-06
Fix static compilation of json2cbor

ffontaine:
json2cbor depends on cjson so tools/json2cbor/json2cbor.o must be
before -lcjson

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>


thiagomacieira:
Good catch! Too bad I've just released v0.4.1...

thiagomacieira:
Trying the "rebase and merge" feature. It won't have my signed-off-by, but let's see what it does.

PR #67, opened 2017-03-14
Chunked string parsing

thiagomacieira:
This is a new v0.5 feature: the ability to parse chunked CBOR strings per chunk. It also adds the ability to do zero-copy parsing of strings. From the docs:

```c
   if (cbor_value_is_text_string(value)) {
       char *ptr;
       size_t len;
       while (1) {
           err = cbor_value_get_text_string(value, &ptr, &len, &value));
           if (err) return err;
           if (ptr == NULL) return CborNoError;
           consume(ptr, len);
       }
   }
```

There are two commits in this series:

1. Add support for parsing chunked strings
2. rewrite the iterate_string_chunks function with new chunk code

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -1104,60 +1104,34 @@ static CborError iterate_string_chunks(const CborValue *value, char *buffer, siz
 {
     cbor_assert(cbor_value_is_byte_string(value) || cbor_value_is_text_string(value));
 
-    size_t total;
     CborError err;
-    const uint8_t *ptr = value->ptr;
-    if (cbor_value_is_length_known(value)) {
-        /* easy case: fixed length */
-        err = extract_length(value->parser, &ptr, &total);
+    CborValue tmp;
+    size_t total = 0;
+    const void *ptr;
+
+    if (!next)
+        next = &tmp;
+    *next = *value;
+    *result = true;
+
+    while (1) {
+        size_t newTotal;
+        size_t chunkLen;
+        err = get_string_chunk(next, &ptr, &chunkLen);
         if (err)
             return err;
-        if (total > (size_t)(value->parser->end - ptr))
-            return CborErrorUnexpectedEOF;
-        if (total <= *buflen)
-            *result = !!func(buffer, ptr, total);
+        if (!ptr)
+            break;
+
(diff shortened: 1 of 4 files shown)


thiagomacieira:
@lpereira @otaviobp @rzr @erichkeane @alradmsft

src/cborparser.c
```diff
@@ -932,14 +932,157 @@ CborError cbor_value_get_int_checked(const CborValue *value, int *result)
...
+ * cbor_value_copy_text_string() does.
+ *
+ * It is designed to be used in code like:
+ *
+ * \code
+ *   if (cbor_value_is_text_string(value)) {
+ *       char *ptr;
+ *       size_t len;
+ *       while (1) {
+ *           err = cbor_value_get_text_string(value, &ptr, &len, &value));
```
otaviobp: There is no cbor_value_get_text_string defined here. Am I missing something or this should be cbor_value_get_text_string_chunk?

src/cborparser.c
```diff
@@ -932,14 +932,157 @@ CborError cbor_value_get_int_checked(const CborValue *value, int *result)
...
+ * cbor_value_copy_byte_string() does.
+ *
+ * It is designed to be used in code like:
+ *
+ * \code
+ *   if (cbor_value_is_byte_string(value)) {
+ *       char *ptr;
+ *       size_t len;
+ *       while (1) {
+ *           err = cbor_value_get_byte_string(value, &ptr, &len, &value));
```
otaviobp: same here

src/cborparser.c
thiagomacieira: cbor_value_get_xxx_string is an inline function in cbor.h which asserts that the type is correct, then calls _cbor_value_get_string. This documentation (and below) apply to the public function.

otaviobp: The inline function is called cbor_value_get_xxxx_string_chunk. I couldn't find a _cbor_value_get_string too. Unless there is a macro somewhere defining them, there is a problem with names in the sample code.

thiagomacieira: You're absolutely right. Thanks for catching this.

thiagomacieira:
With no reply, assuming everything is ok.

PR #68, opened 2017-03-29
json2cbor: Initialize index before "for scope"

rzr:
Observed issue on earlier version:

  json2cbor.c:157:5: \
  error: for loop initial declarations are only allowed in C99 mode

Note, it's not manatory for current master branch
since build script already sets the c99 flag,
but it won't cause any harm.

Signed-off-by: Philippe Coval <philippe.coval@osg.samsung.com>

### DIFF
--- a/tools/json2cbor/json2cbor.c
+++ b/tools/json2cbor/json2cbor.c
@@ -149,12 +149,13 @@ uint8_t *decode_base64url(const char *string, size_t *len)
 
 uint8_t *decode_base16(const char *string, size_t *len)
 {
+    size_t i;
     *len = strlen(string) / 2;
     uint8_t *buffer = malloc(*len);
     if (buffer == NULL)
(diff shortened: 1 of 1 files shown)


tools/json2cbor/json2cbor.c
```diff
@@ -154,7 +154,8 @@ uint8_t *decode_base16(const char *string, size_t *len)
     if (buffer == NULL)
         return NULL;
 
-    for (size_t i = 0; i < *len; ++i) {
+    size_t i;
```
thiagomacieira: Well, mixing variable declarations and statements is also only allowed in C99. If you're going to try and solve the warning, move the declaration to the beginning of the function.

thiagomacieira:
Overriding the CI and approving. Travis is broken and I haven't configured AppVeyor for master branch.

PR #71, opened 2017-06-09
Hide the API using FILE* from freestanding C implementations

thiagomacieira:


thiagomacieira:
@lpereira

PR #74, opened 2017-07-26
Provide shared library

CoRfr:
Some people may want to use tinycbor as a shared library instead of
as a static library, so providing a set of .so accordingly.

Even with the static library, I'm having trouble incorporating it
into another shared library, so providing -fPIC option as well.

Signed-off-by: Bertrand Roussel <broussel@sierrawireless.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,9 @@ CBORDUMP_SOURCES = tools/cbordump/cbordump.c
 
 INSTALL_TARGETS += $(bindir)/cbordump
 INSTALL_TARGETS += $(libdir)/libtinycbor.a
+INSTALL_TARGETS += $(libdir)/libtinycbor.so
+INSTALL_TARGETS += $(libdir)/libtinycbor.so.0
+INSTALL_TARGETS += $(libdir)/libtinycbor.so.$(VERSION)
 INSTALL_TARGETS += $(pkgconfigdir)/tinycbor.pc
 INSTALL_TARGETS += $(TINYCBOR_HEADERS:src/%=$(includedir)/tinycbor/%)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ libdir = $(exec_prefix)/lib
 includedir = $(prefix)/include
 pkgconfigdir = $(libdir)/pkgconfig
 
-CFLAGS = -Wall -Wextra
+CFLAGS = -Wall -Wextra -fPIC
```
thiagomacieira: Sorry, can't have that in the default flags.

CoRfr:
I was not aware that the -fPIC would make the code bigger. The new patch doesn't touch the generation of the .a. It does however add some .pic.o built from the same .c files but with -fPIC.

I compared `lib/libtinycbor.a` and `bin/cbordump` from before & after the build and they are the same (a `cmp` between before/after doesn't show any difference).

I can still change the destination branch if you prefer.

thiagomacieira:
Yes, it's still a new feature, so please retarget to dev.

TinyCBOR doesn't have much in terms of static data that would cause it to require PIC registers, but some older compilers are stupid and may reserve the register on x86 even if not needing it.

thiagomacieira:
One more thing: I am not yet promising binary compatibility. In fact, I have a change I'd like to make for 0.6 that will break it.

CoRfr:
Sounds good for dev, I just changed the base.

CoRfr:
@thiagomacieira, could you please review the latest changes?
- files compiled with -fPIC have a .pic.o extension and are not used at all in the static compilation path
- base branch is 'dev'

thiagomacieira:
Apologies for the delay. I've been thinking all week whether I should just switch the buildsystem to CMake.

I haven't come to a conclusion. So I'll take your patch as-is for now. It's not fair to you to keep you waiting.

PR #78, opened 2017-11-13
Make sure that -fPIC objects are removed in make clean

thiagomacieira:
We were leaving the .pic.o files behind

@CoRfr


thiagomacieira:
Ugh... Bad move, this commit had a spelling mistake. I got confused somewhere and pushed what I shouldn't.

PR #79, opened 2017-11-13
Fix MSVC build: cborvalidation.obj wasn't added to the library

thiagomacieira:
My own fault....


thiagomacieira:
Self-approving to fix a build failure.

PR #82, opened 2017-11-14
Fix build on non-ELF systems (macOS): don't build the shared library

thiagomacieira:
Detect if this is an ELF system by checking if /bin/sh is ELF. This can
be configured manually with:
   make BUILD_SHARED=0
or
   make BUILD_SHARED=1

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,9 +33,11 @@ TINYCBOR_SOURCES = \
 #
 CBORDUMP_SOURCES = tools/cbordump/cbordump.c
 
+BUILD_SHARED = $(shell file -L /bin/sh 2>/dev/null | grep -q ELF && echo 1)
+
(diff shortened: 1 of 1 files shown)


thiagomacieira:
@CoRfr

Makefile
```diff
@@ -33,9 +33,11 @@ TINYCBOR_SOURCES = \
 #
 CBORDUMP_SOURCES = tools/cbordump/cbordump.c
 
+BUILD_SHARED = $(shell file -L /bin/sh 2>/dev/null | grep -q ELF && echo 1)
```
erichkeane: Would this override the 'make' line, or will the build system ignore this line?

thiagomacieira: It's the other way around: the command-line overrides Makefile assignments.

PR #89, opened 2017-12-18
Encoder: Fix the returning of OOM errors in closing containers

thiagomacieira:
Conditional added by 3369dd1f017598f74c3f6f36591505377d72bd9d was
inverted.


thiagomacieira:
Let me look into adding a test. Shouldn't be too difficult.

thiagomacieira:
The test returned the error, but it OR'ed with the previous error, so we weren't catching whether the close_container function did return OOM.

PR #94, opened 2017-12-21
Pretty: limit the number of recursions, like the Parser

thiagomacieira:
The pretty-dumper code did not have a limit, assuming that the input had
been validated beforehand, so if the nesting was too deep, one would
have got an error before getting to the dumper.

However, it's not inconceivable that someone adds a logger code to a
packet a bit too early. So let's be defensive and apply the limiter here
too.

### DIFF
--- a/src/cborinternal_p.h
+++ b/src/cborinternal_p.h
@@ -31,6 +31,10 @@
 #  define CBOR_INTERNAL_API
 #endif
 
+#ifndef CBOR_PARSER_MAX_RECURSIONS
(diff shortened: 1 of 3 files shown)


erichkeane:
+1 LGTM

lpereira:
Nice to see this -- I imagine cbordump isn't crashing any longer with the files I provided you?  (It takes a few hours to check for myself.)

thiagomacieira:
I didn't actually try your files... But I'll try.

PR #99, opened 2017-12-26
Parser: merge the break-handling code in enter_container()

thiagomacieira:
There was a section of code that was mostly the same, dealing with the
possibility of reading a Break (0xFF) as the next element in a map or
array. The only difference was that preparse_next_value() also
decremented the item count for the current list.

We could have used that, but that would have meant a +1 on the list,
which in turn lowers the number of possible elements from UINT32_MAX-1
to UINT32_MAX-2. Not a big deal, but we can avoid it by just splitting
to a new function and retaining tail-call optimisations.

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -297,15 +297,9 @@ static CborError preparse_value(CborValue *it)
     return CborNoError;
 }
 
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Updated.

PR #100, opened 2017-12-27
Valiation: suppress warning about uninitialized variables

thiagomacieira:
GCC7 complains, though it's wrong:
```
 src/cborvalidation.c:472:57: error: ‘previous’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
 src/cborvalidation.c:485:61: error: ‘previous_end’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
```

### DIFF
--- a/src/cborvalidation.c
+++ b/src/cborvalidation.c
@@ -436,16 +436,15 @@ static CborError validate_container(CborValue *it, int containerType, int flags,
 {
     CborError err;
     const uint8_t *previous = NULL;
-    const uint8_t *previous_end;
+    const uint8_t *previous_end = NULL;
 
     if (!recursionLeft)
         return CborErrorNestingTooDeep;
 
     while (!cbor_value_at_end(it)) {
-        const uint8_t *current;
+        const uint8_t *current = cbor_value_get_next_byte(it);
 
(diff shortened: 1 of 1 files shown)


src/cborvalidation.c
```diff
@@ -436,16 +436,15 @@ static CborError validate_container(CborValue *it, int containerType, int flags,
 {
     CborError err;
     const uint8_t *previous = NULL;
-    const uint8_t *previous_end;
+    const uint8_t *previous_end = NULL;
```
lpereira: This silences the warning, but does it really fix the problem?  This is essentially transforming the `size_t bytelen1 = (size_t)(previous_end - previous)` statement into `size_t bytelen1 = (size_t)(0 - previous)` in the case where previous_end wouldn't be initialized.

thiagomacieira: There was no problem to begin with. Both uses of those variables happen after
```c++
    if (containerType != CborMapType)
        continue;
```
[line 467]

That means `containerType` was `CborMapType`, which means `current` **is** properly intialized. In turn, that means `previous` will be properly initialized for the next iteration.

Issue #102, opened 2018-01-02
tinycbor does not link on Windows 7 using MinGW.org GCC-6.3.0-1


telkamp:
Building tinycbor on my Windows 7 machine using MinGW GCC-6.3.0-1 (g++ compiler) caused a linker error:
lib/tinycbor/libtinycbor.a(cborencoder.c.obj): In function create_container:
cborencoder.c:460: undefined reference to static_assert
cborencoder.c:461: undefined reference to static_assert
collect2.exe: error: ld returned 1 exit status

It seems to me that your #if term in compilersupport_p.h(53) is not working in my case:
`#if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_assert >= 200410`
`#  define cbor_static_assert(x)         static_assert(x, #x)`
`#elif !defined(__cplusplus) && defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 406) && (__STDC_VERSION__ > 199901L)`
`#  define cbor_static_assert(x)         _Static_assert(x, #x)`
`#else`
`#  define cbor_static_assert(x)         ((void)sizeof(char[2*!!(x) - 1]))`
`#endif`

The first term `__STDC_VERSION__ >= 201112L` is true in my case, but I don't have the static_assert.
Can the first term be removed?
Otherwise: I found a quite similar term used in the RapidJSON library:
https://github.com/Tencent/rapidjson/blob/a7f687fdf812b9f9c7bf7f78eeb1261642f31623/include/rapidjson/rapidjson.h#L419
`#if __cplusplus >= 201103L || ( defined(_MSC_VER) && _MSC_VER >= 1800 )`
`#define RAPIDJSON_STATIC_ASSERT(x) \`
`   static_assert(x, RAPIDJSON_STRINGIFY(x))`
`#endif // C++11`
This might be an alternative.

thiagomacieira:
If {{__STDC_VERSION__}} is 201112 and static_assert is failing, your compiler is broken.

Check if assert.h has a macro definition for {{static_assert}}. That's probably the issue. Please report upstream. Meanwhile, compile TinyCBOR in C99 mode (which is the default).

thiagomacieira:
I doubt this is a TinyCBOR bug.

telkamp:
My include/assert.h does not include a static_assert macro
I've re-installed MinGW completely, but this does not help. Compiler version is still mingw32-gcc-g++.
[...]

PR #103, opened 2018-01-05
static_assert Support for MinGW32 in C11 mode

telkamp:
Workaround for missing static_assert macro in MinGW32's glibc.
 See C11 standard, chapter 7.2 "Diagnostics <assert.h>", paragraph 3) */
See [https://github.com/intel/tinycbor/issues/102](https://github.com/intel/tinycbor/issues/102)

### DIFF
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -50,6 +50,14 @@
 #  include <immintrin.h>
 #endif
 
+/* workaround for missing static_assert macro in MinGW32's glibc 
+ * See C11 standard, chapter 7.2 "Diagnostics <assert.h>", paragraph 3) */
+#if !defined (__cplusplus) && defined(__MINGW32__) && !defined(__MINGW64__) \
+    && (defined (_ISOC11_SOURCE) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)))
+#  undef  static_assert
+#  define static_assert                 _Static_assert
+#endif
+
 #if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_assert >= 200410
 #  define cbor_static_assert(x)         static_assert(x, #x)
 #elif !defined(__cplusplus) && defined(__GNUC__) && (__GNUC__ * 100 + __GNUC_MINOR__ >= 406) && (__STDC_VERSION__ > 199901L)


thiagomacieira:
Didn't @erichkeane's suggestion work?

telkamp:
I've not tested the line from @erichkaene, so it might work. But I would prefer a solution handling specifically the MinGW case to avoid any side effects to other configurations as much as possible. (@erichkaene's solution will effect non-MinGW cases as well)
So my patch does only what MinGW's glibc should have already done and nothing more - it defines the `static_assert` macro. Please have in mind that MinGW64 defines the `static_assert`.
I will send you another pull request with my Signed-off-by line.

erichkeane:
That seems bad to change the static assert definition rather than just correctly support the cbor assert.  
Also, if you're going to mix || and && like that, ad parens.

telkamp:
I'm really not sure if the current cbor assert is faulty or correct, so I would still prefer to add a #define only for what was missing in MinGW's glibc.
The parens are correct.
But of course, it is only my proposal. If you make an alternative pull request, I will test it and report if it works or not.

erichkeane:
The parens are not, there is still an && and a || in the same statement without them, this is a warning on newer compilers.

Also, this patch isn't very future proof, in the case where glibc fixes this...  Also, isn't it a function macro, why does this patch have it as a keyword macro? 

From a different perspective, I REALLY hate the idea of using compiler preprocessor defines to identify a library.  There's no reason someone couldn't simply use a different glibc with mingw that would end up hitting this.

telkamp:
Maybe I'm misunderstanding something here - as I can see, all && and || terms are separated with parens here:
`#if !defined(__cplusplus) && defined(__MINGW32__) && !defined(__MINGW64__) 
    && (defined (_ISOC11_SOURCE) || (defined (__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)))` 

Furthermore, the patch **is** future proof - that's why there is a `#undef` before eventually re-defining `static_assert`. (Please have in mind that the `#if` statement ensures that we have C11, so we know that `static_assert` could eventually have been defined by a macro before, if glibc really fixed it some day).
And it is a keyword macro and not a function macro, because glibc defines it as keyword macro also. Please have a look to the original code: [https://sourceware.org/git/?p=glibc.git;a=blob;f=assert/assert.h;hb=HEAD#l138]()

But I agree with you that checking for libraries with preprocessor defines is really a bad solution - but much better than a non-working solution. But if you know something better, I would be happy to learn something new.

erichkeane:
The ISOC11 and the STDC checks aren't. 

I DID provide a better working solution in the last thread, which was to only use static_assert if it is defined.

telkamp:
So you can close my pull request without merging it.
I will give a feedback to the already closed issue https://github.com/intel/tinycbor/issues/102 if an new version of tinycbor will compile on my environment without modifications.

thiagomacieira:
In other words, in C++11 or later, use `static_assert`; in C11, use `_Static_assert`; with a poor man's fallback.

telkamp:
I think that's a gxood solution. I will currently not be able to provide another patch, but I could provide a feature request following your proposal. What do you think?

PR #108, opened 2018-01-27
Parser: remove the cbor_value_get_xxxx_string_chunk() API

thiagomacieira:
It's really good, but I need to redesign it in order to support the
chained buffers found in Zephyr[1][2] and Mynewt. I'll bring it back in
0.6 when I introduce support for parsing directly from chained buffers
too.

[1] http://docs.zephyrproject.org/api/networking.html#network-buffers
[2] https://github.com/zephyrproject-rtos/zephyr/blob/master/include/net/buf.h


thiagomacieira:
@georgen117: it worked. can you look into this one too?

src/cbor.h
```diff
@@ -452,21 +452,6 @@ CBOR_INLINE_API CborError cbor_value_dup_byte_string(const CborValue *value, uin
...
-    assert(cbor_value_is_text_string(value));
-    return _cbor_value_get_string_chunk(value, (const void **)bufferptr, len, next);
-}
-CBOR_INLINE_API CborError cbor_value_get_byte_string_chunk(const CborValue *value, const uint8_t **bufferptr,
-                                                           size_t *len, CborValue *next)
-{
-    assert(cbor_value_is_byte_string(value));
-    return _cbor_value_get_string_chunk(value, (const void **)bufferptr, len, next);
-}
-
```
georgen117: Right now `_cbor_value_get_string_chunk`, `cbor_value_get_text_string_chunk`, and `cbore_value_get_byte_string_chunk` look like they are all in a public header have they been part of a release yet?  Is it possible that they are already used by other projects?

That is my major concern when removing the headers.

I can see that  `_cbor_value_get_string_chunk` was already moved to cborinternal_p.h which is probably where is should have been to start with.

src/cborparser.c
```diff
@@ -1040,87 +1040,7 @@ static CborError get_string_chunk(CborValue *it, const void **bufferptr, size_t
...
- * The \a next pointer, if not null, will be updated to point to the next item
- * after this string. During iteration, the pointer must only be passed back
- * again to this function; passing it to any other function in this library
- * results in undefined behavior. If there are no more chunks to be read from
- * \a value, then \a next will be set to the next item after this string; if \a
- * value points to the last item, then \a next will be invalid.
- *
- * \sa cbor_value_dup_byte_string(), cbor_value_copy_byte_string(), cbor_value_caculate_string_length(), cbor_value_get_text_string_chunk()
- */
-
```
georgen117: The commit message stated that this was going to be reintroduced in a future commit.  This documentation is great quality are you going to need to rewrite when reintroduced or is the plan to use this documentation in a future update?

src/cbor.h
thiagomacieira: Those functions are new in 0.5. I'm removing them so I can redesign them for when we're reading from something like Zephyr's `net_buf`.

src/cborparser.c
thiagomacieira: I have a commit reverting this right after the version number change for the dev branch. I'll update the docs as I figure out what the API will look like.

thiagomacieira:
Thanks. I have a minor change now and then I'll release 0.5.

georgen117:
Your welcome. Sorry I took so long. I am still learning the code base so I wanted to take my time to understand what was being changed.

thiagomacieira:
Oh, thanks for all the help. It's much appreciated, even if all you do is see that it's syntactically correct C code ;-)

The dev changes are a little trickier.

PR #109, opened 2018-01-30
Make it easier to build only the freestanding code

thiagomacieira:


thiagomacieira:
Fair question. I've updated the commit message with more details.

_Freestanding implementation_ is defined by the C standard, clause 4 (Conformance), paragraph 6.

PR #114, opened 2018-02-05
json2cbor: don't crash if realloc() returns false

thiagomacieira:
perror() doesn't exist. It's not like Perl's die().

### DIFF
--- a/tools/json2cbor/json2cbor.c
+++ b/tools/json2cbor/json2cbor.c
@@ -448,8 +448,10 @@ int main(int argc, char **argv)
         buffer = NULL;
         do {    // it the hard way
             buffer = realloc(buffer, buffersize + chunk);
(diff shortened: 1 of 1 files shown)


tools/json2cbor/json2cbor.c
```diff
@@ -448,8 +448,10 @@ int main(int argc, char **argv)
         buffer = NULL;
         do {    // it the hard way
             buffer = realloc(buffer, buffersize + chunk);
-            if (buffer == NULL)
+            if (buffer == NULL) {
                 perror("malloc");
```
erichkeane: Not sure giving perror a string is valuable (since the message becomes "malloc:Out of memory", but this is existing.

ediosyncratic:
Typo in commit message: s/exist/exit/ !  Too late, I know ...

PR #115, opened 2018-02-05
Docs comments updates

thiagomacieira:


tools/json2cbor/json2cbor.c
```diff
@@ -365,7 +365,7 @@ CborError decode_json(cJSON *json, CborEncoder *encoder)
 
         for (item = json->child ; item; item = item->next) {
             if (usingMetaData && strlen(item->string) > strlen(meta_data_marker)
-                    && strcmp(item->string + strlen(item->string) - 5, meta_data_marker) == 0)
+                    && strcmp(item->string + strlen(item->string) - strlen(meta_data_marker), meta_data_marker) == 0)
```
erichkeane: Should we prefer sizeof -1 here?  it IS a constant string...  Could be done above too.

Either way, calculating this 2x in this ifstmt is kinda crappy.

thiagomacieira: strlen(constant) is usually replaced by the compiler with a constant. You of all people should know that :-)

src/cbor.dox
```diff
@@ -80,10 +80,10 @@
  * \value CborArrayType             Type is a CBOR array
  * \value CborMapType               Type is a CBOR map (an associative container with key and value pairs)
  * \value CborTagType               Type is a CBOR tag (a 64-bit integer describing the item that follows, see CborKnownTags)
- * \value CborSimpleType            Type is one of CBOR Simple Types
+ * \value CborSimpleType            Type is one of the CBOR Simple Types
  * \value CborBooleanType           Type is a boolean (true or false)
  * \value CborNullType              Type encodes a null
- * \value CborUndefinedType         Type encodes an undefined value
+ * \value CborUndefinedType         Type encodes a undefined value
```
georgen117: shouldn't this still be `an undefined value` is there a rule I am unaware?

georgen117: according to https://www.grammar.com/a-vs-an-when-to-use/

words starting with U use **a** when its pronounced like the word 'you' (a university)

when is starts with vowel sound 'uh' us **an** (an umbrella)

So I think this is `an undefined value` this is minor grammar but most of the changes are minor grammar so I don't know if you want to fix.

thiagomacieira: I changed the wrong line...

PR #116, opened 2018-02-06
Doxygen updates

thiagomacieira:
I think this is the final item from the reviews that happened at the Qt change review https://codereview.qt-project.org/#/c/217413/3/


Doxyfile
```diff
@@ -1,2318 +1,50 @@
...
-# The default value is: YES.
-
-WARNINGS               = YES
-
-# If the WARN_IF_UNDOCUMENTED tag is set to YES, then doxygen will generate
-# warnings for undocumented members. If EXTRACT_ALL is set to YES then this flag
-# will automatically be disabled.
-# The default value is: YES.
-
 WARN_IF_UNDOCUMENTED   = NO
```
georgen117: just wondering why not use the WARN_IF_UNDOCUMENTED tag?

thiagomacieira: Because then it extracts too much. What I left undocumented I probably intended to leave undocumented.

Doxyfile
```diff
@@ -1,2318 +1,50 @@
...
-
-EXCLUDE_SYMBOLS        =
-
-# The EXAMPLE_PATH tag can be used to specify one or more files or directories
-# that contain example code fragments that are included (see the \include
-# command).
-
-EXAMPLE_PATH           = . \
-                         ../examples/rime \
-                         ../examples/multi-threading
```
georgen117: EXAMPLE_PATH was removed.  Since the path points outside the tinycbor project I assume that this was purposely removed.

Doxyfile
```diff
@@ -1,2318 +1,50 @@
...
-# irrespective of the value of the RECURSIVE tag.
-# The default value is: NO.
-
-EXAMPLE_RECURSIVE      = NO
-
-# The IMAGE_PATH tag can be used to specify one or more files or directories
-# that contain images that are to be included in the documentation (see the
-# \image command).
-
 IMAGE_PATH             = pics
```
georgen117: There is no `pics` folder in tinycbor at this time are we leaving this here for future documentation?

Doxyfile
thiagomacieira: Right, but that was intentional as we have no examples.

Doxyfile
thiagomacieira: I didn't know if this was important. Since the comments had been stripped, I thought it was the **output** dir where doxygen stores images it creates....

georgen117: IMAGE_PATH is used by doxygen to include external images in your documentation. You are not using any external images that I was able to see.  This will not be used unless you use the `\image` tag somewhere in you documentation. I so far have not seen any `\image`  tags. Since its currently not used I would be inclined to remove the tag but leaving it will be harmless.

PR #117, opened 2018-02-06
Validator: simplify (and correct) the map sorting verification

thiagomacieira:
We don't need to compare the map lengths directly. The memcmp is
sufficient, since the source data is big endian.

Of course, verifying for sorting requires the map has known length.


thiagomacieira:
Note: I am actually unsure whether my interpretation of RFC 7049 section 3.10 is correct.... This commit is keeping the current behaviour.

Issue #120, opened 2018-03-13
Cannot build json2cbor


kyegupov:
CentOS 6

**Commands**
```
git clone https://github.com/DaveGamble/cJSON.git
cd cJSON
mkdir build
cd build
cmake ..
make && sudo make install
cd ../..
git clone https://github.com/intel/tinycbor.git
cd tinycbor
make
```
**Error**
```
cc  -I./src -std=c99 -Wall -Wextra  -c -o tools/json2cbor/json2cbor.o tools/json2cbor/json2cbor.c
tools/json2cbor/json2cbor.c:50: warning: declaration does not declare anything
tools/json2cbor/json2cbor.c: In function ‘parse_meta_data’:
tools/json2cbor/json2cbor.c:209: warning: braces around scalar initializer
tools/json2cbor/json2cbor.c:209: warning: (near initialization for ‘result.t’)
tools/json2cbor/json2cbor.c:209: error: incompatible types when initializing type ‘CborType’ using type ‘void *’
tools/json2cbor/json2cbor.c:209: warning: excess elements in struct initializer
tools/json2cbor/json2cbor.c:209: warning: (near initialization for ‘result’)
tools/json2cbor/json2cbor.c:223: error: ‘struct MetaData’ has no member named ‘simpleType’
tools/json2cbor/json2cbor.c:225: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c: In function ‘decode_json_with_metadata’:
tools/json2cbor/json2cbor.c:238: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:239: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:240: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:268: error: ‘struct MetaData’ has no member named ‘simpleType’
tools/json2cbor/json2cbor.c:278: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:280: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:282: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:284: error: ‘struct MetaData’ has no member named ‘v’
tools/json2cbor/json2cbor.c:287: error: ‘struct MetaData’ has no member named ‘v’
[...]

PR #121, opened 2018-03-18
Update gcc flags to enable building json2cbor with gcc 4.4.7 (standard in CentOS 6)

kyegupov:
Fixes https://github.com/intel/tinycbor/issues/120


kyegupov:
I didn't get any extra email, but the commit is amended now.

PR #123, opened 2018-03-30
Incomplete input results in invalid assert

marius-feldmann:
Currently, tinycbor does not allow to hand over incomplete numeric Cbor values. Due to the assert checking for an error while extracting a number from the input, the parsing process globally fails. Instead of checking the error via assert, it should be propagated back to the caller.

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -326,7 +326,9 @@ static CborError advance_internal(CborValue *it)
 {
     uint64_t length;
     CborError err = _cbor_value_extract_number(&it->ptr, it->parser->end, &length);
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Can you show some code of how this could happen? I need to add a unit test.

marius-feldmann:
Actually, we discovered, that there is a bug is on our side. We did not properly check if we have reached EOF (more precisely, we ignored CborErrorUnexpectedEOF) and continued parsing, thus, ran into the assert. I close the pull request - there is no issue with tinycbor. Sorry for not analysing our code deeply enough.

PR #124, opened 2018-05-08
Attempt at detecting a few more endiannesses

thiagomacieira:
On INTEGRITY, the __BYTE_ORDER__ macro isn't defined. So fall back to
the processor macros.


thiagomacieira:
Tested in Qt's INTEGRITY machine and worked, so self-approving.

PR #127, opened 2018-05-24
validation: Fix tag array for 8 bit architectures

bergzand:
Compilation for 8 bit architectures fails due to a bit shift larger than 8 for a uint8_t data type. This PR changes the data type to uint16_t and uint32_t for these cases.


bergzand:
@thiagomacieira I'm working on integrating tinycbor as an external library for RIOT-OS (see RIOT-OS/RIOT#9020). Our CI spotted the overflow compilation issues when compiling for the AVR (8-bit) and MSP430 (16-bit) MCU's. I plan on testing some more tomorrow with some arduino boards.

> Your change is correct, but please update the src/parsetags.pl script to generate the new code. Use uint32_t for everything.

Thank you for your feedback. I've added two commits to reflect these changes. Please let me know if you want me to squash out/remove the first commit of this PR.

bergzand:
@thiagomacieira A quick test on an arduino uno board shows that at least the parser is correctly functioning. I have to enable the `CBOR_NO_FLOATING_POINT` though because `FP_NAN`, `FP_INFINITE` and `fpclassify` are not defined.

thiagomacieira:
Thanks. I'd prefer squashed into a single commit.

thiagomacieira:
That also reminds me: I need your commits to contain a Signed-off-by line. I forgot in the other two I merged because they didn't require any updates, but please update this one.

bergzand:
@thiagomacieira Squashed and now with a signed-off-by line. Thanks again for the feedback

thiagomacieira:
Thank you Koen.

PR #130, opened 2018-05-25
Add more byte order detection code

thiagomacieira:
Suggested in https://sourceforge.net/p/predef/wiki/Endianness/ and also
found by looking into glibc's <endian.h>. Plus some recommendation from
other people.

This fixed the build on INTEGRITY.


thiagomacieira:
Self-approving based on the fact that this was needed to get INTEGRITY to build.

PR #138, opened 2018-09-16
Fix #137: off-by-one error in UTF-8 decoding

### LINKED ISSUE #137: buffer bounds check error in get_utf8 function
Consider the following code snippet:

```
int main(void)
{
    // NB! The 2-nd byte in buffer is a canary
    const uint8_t buffer[] = {0xC0 | 0x1F, 0x80 | 0x3F};

    const uint8_t *begin = buffer;
    // NB! data size provided to get_utf8 function is 1 byte
    uint32_t result = get_utf8(&begin,  begin + 1);

    printf("buffer ptr: %p, begin ptr: %p, return value: %x\n",
        buffer, begin, result);

    return 0;
}
```

the actual output is:

> buffer ptr: 0x7ffd07e9d700, begin ptr: 0x7ffd07e9d702, return value: 7ff

the expected output is:

>buffer ptr: 0x7ffd07e9d700, begin ptr: 0x7ffd07e9d701, return value: ffffffff

Solution:
```
diff --git a/src/utf8_p.h b/src/utf8_p.h
index 577e540..ca43835 100644
--- a/src/utf8_p.h
+++ b/src/utf8_p.h
@@ -66,7 +66,7 @@ static inline uint32_t get_utf8(const uint8_t **buffer, const uint8_t *end)
         return ~0U;
     }
 
-    if (n < charsNeeded - 1)
+    if (n < charsNeeded)
         return ~0U;
 
     /* first continuation character */
```

Have a nice day!

thiagomacieira:
I ported this from code I had written for Qt, where the subtraction was
correct. Qt's code is roughly:
```
    uchar b = *src++;
    ...
    int bytesAvailable = Traits::availableBytes(src, end);
    if (Q_UNLIKELY(bytesAvailable < charsNeeded - 1)) {
```

Which means src was advanced by the time we call Traits::availableBytes,
meaning that call returns the number of continuation bytes only. Our
code was:
```
    ptrdiff_t n = end - *buffer;
    ...
    uc = *(*buffer)++;
    ...
    if (n < charsNeeded - 1)
```

Which means our n is the total number of bytes, including the first
byte.

### DIFF
--- a/src/utf8_p.h
+++ b/src/utf8_p.h
@@ -66,7 +66,7 @@ static inline uint32_t get_utf8(const uint8_t **buffer, const uint8_t *end)
         return ~0U;
     }
 
-    if (n < charsNeeded - 1)
+    if (n < charsNeeded)
         return ~0U;
 
     /* first continuation character */
--- a/tests/parser/tst_parser.cpp
+++ b/tests/parser/tst_parser.cpp
@@ -1762,24 +1762,36 @@ void tst_Parser::strictValidation_data()
     QTest::newRow("overlong-_stringx2-0*8") << raw("\x7f\x60\x7b\0\0\0\0\0\0\0\0\xff") << int(CborValidateShortestNumbers) << CborErrorOverlongEncoding;
 
     // strict mode
-    QTest::newRow("invalid-utf8-1char") << raw("\x61\x80") << int(CborValidateStrictMode) << CborErrorInvalidUtf8TextString;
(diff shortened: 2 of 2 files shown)


thiagomacieira:
Let me see if I can add a unit test that replicates the situation from the bug report...

georgen117:
I in inadvertently used my other account to review this change just making sure you knew it was me reviewing. "gnash117" and "georgen117" same person different accounts.

The change looks good.

thiagomacieira:
Took me a while, but I got the tests added.

PR #148, opened 2018-12-05
Decrease unittests verbosity

phirsov:
Make tests less verbose to avoid Travis log cluttering with test progress messages


thiagomacieira:
If you have a strong reason why verbosity needs to be reduced, please say so.

phirsov:
In my vision, Travis purpose is to check whether the commited changes break code or not. Travis log is cluttered with unittest messages too much (5k lines), while the result of the test suit run is binary: OK of FAIL. If one needs the detailed view of failed and passed tests, he (or she) can run tests by hand.

thiagomacieira:
Except when you can't reproduce the error on your own. It's good to know what failed.

Anyway, there's a middle-ground solution: pass the extra make parameter `TESTARGS=-silent`. The QtTest runner will then only print the failures, but not the successes.

phirsov:
Tweak, TESTARGS=-silent is used now

PR #151, opened 2018-12-08
Use Qt version provided by Trusty Tahr (Debian Jessie) out-of-the-box

phirsov:
Some people are suspicious about using 3rd-party repositories (PPA's) in their projects. The goal of this PR is to build project with Qt version provided by Ubuntu (Debian) LTS releases repositories out-of-the-box.

### DIFF
--- a/src/cborpretty.c
+++ b/src/cborpretty.c
@@ -526,7 +526,10 @@ static CborError value_to_pretty(CborStreamFunction stream, void *out, CborValue
             err = stream(out, "%s%" PRIu64 ".%s", val < 0 ? "-" : "", ival, suffix);
         } else {
             /* this number is definitely not a 64-bit integer */
-            err = stream(out, "%." DBL_DECIMAL_DIG_STR "g%s", val, suffix);
+            if (fpclassify(val) == FP_NAN)
+                err = stream(out, "nan");
+            else
+                err = stream(out, "%." DBL_DECIMAL_DIG_STR "g%s", val, suffix);
         }
         break;
     }
--- a/tests/parser/tst_parser.cpp
+++ b/tests/parser/tst_parser.cpp
@@ -104,10 +104,30 @@ private slots:
 static CborError qstring_printf(void *out, const char *fmt, ...)
 {
     auto str = static_cast<QString *>(out);
+
     va_list va;
+
     va_start(va, fmt);
-    *str += QString::vasprintf(fmt, va);
+    char c;
+    int buf_size = vsnprintf(&c, 1, fmt, va) + 1;
     va_end(va);
+
+    if (buf_size < 0)
+        return CborErrorIO;
+
+    else if (buf_size == 0)
+        return CborNoError;
+
+    va_start(va, fmt);
+    QByteArray buf(buf_size, '\0');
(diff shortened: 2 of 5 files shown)


src/cborpretty.c
```diff
@@ -526,7 +526,10 @@ static CborError value_to_pretty(CborStreamFunction stream, void *out, CborValue
             err = stream(out, "%s%" PRIu64 ".%s", val < 0 ? "-" : "", ival, suffix);
         } else {
             /* this number is definitely not a 64-bit integer */
-            err = stream(out, "%." DBL_DECIMAL_DIG_STR "g%s", val, suffix);
+            if (fpclassify(val) == FP_NAN)
```
thiagomacieira: Please explain why this is needed.

tests/parser/tst_parser.cpp
```diff
@@ -112,10 +112,10 @@ static CborError qstring_printf(void *out, const char *fmt, ...)
     int buf_size = vsnprintf(&c, 1, fmt, va) + 1;
     va_end(va);
 
-    if (buf_size < 0)
+    if (buf_size <= 0)
```
thiagomacieira: This is wrong. vsnprintf returns negative on error; zero is not an error.

erichkeane: Well, in this case you're already doing a +1 on line 112.  A buf_size of 0 means vsnprintf returned -1.

phirsov:
Trusty Tahr is still supported until April 2019. Moreover, this release is used in Travis CI system (that's why 3-rd party PPA is needed in .travis.yml to pass the check). The effort needed for backward compatibility is minimal.

P.S. I don't exactly know if the Qt version provided by more recent Ubuntu LTS release Xenial Xerus repo is sufficient to pass the check.

src/cborpretty.c
phirsov: As you can see in [AppVeyor build](https://ci.appveyor.com/project/phirsov/tinycbor/builds/20867206/job/eladqc8ebdn56gnc) Visual Studio pretty-printing of NaN is somewhat weird

thiagomacieira:
I don't care how long Ubuntu is supported. As of January 2019, the oldest Qt version supported by Qt is 5.9.7. In fact, I'd prefer moving all Linux builds to Xenial and still use Qt 5.9.

tests/parser/tst_parser.cpp
thiagomacieira: Oh, right. Good point.

src/cborpretty.c
thiagomacieira: I see it now, but it was working  before. What changed?

In any case, please err on the side of writing more in the commit message.

phirsov: QString::vasprintf was replaced with pure-C vsnprintf in  tests/parser/tst_parser.cpp, because this member function was introduced in the newer Qt version. vsnprintf NaN pretty-printing is broken in Visual Studio.

phirsov:
OK, let's abandon this PR. I keep this fork in my own repo alone with other fork, backward compatible with ancient Qt4. Everyone, who need, can use it.

phirsov:
Refused by maintainer.

PR #158, opened 2019-04-04
Tests: Catch an earlier QCOMPARE failure in compare()

thiagomacieira:


lpereira:
Looks harmless enough, but what's the story here?

thiagomacieira:
> Looks harmless enough, but what's the story here?

`QCOMPARE` macro has a return in case of failure. If a test fails inside the `encodeOne` function, we log that error, but were proceeding to perform more tests (which could fail again and produce more errors).

lpereira:
That would look nice in the commit message :)

+1 for the patch

thiagomacieira:
> That would look nice in the commit message :) +1 for the patch

Ok. Sorry, it was obvious to me :-)

PR #159, opened 2019-06-05
Fix compilation error in the documentation

iamsergio:


thiagomacieira:
Strictly speaking, it's not wrong. The address of an array is the address of its first element. It's not a pointer.

My guess is that your code was not quite like the example.

Anyway, accepting.

PR #164, opened 2019-10-09
Update cborencoder.c documentation

elie-elkhoury:
Fixed the error checking example proposed in the comments/documentation at the beginning of the file.
Changed occurrences of `if(!err)` to `if(err)`, and `cbor_assert(err)` to `cbor_assert(!err)`.

### DIFF
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -117,16 +117,16 @@
  *      CborEncoder encoder, mapEncoder;
  *      cbor_encoder_init(&encoder, buf, sizeof(buf), 0);
  *      err = cbor_encoder_create_map(&encoder, &mapEncoder, 1);
- *      if (!err)
+ *      if (err)
  *          return err;
  *      err = cbor_encode_text_stringz(&mapEncoder, "foo");
- *      if (!err)
+ *      if (err)
  *          return err;
  *      err = cbor_encode_boolean(&mapEncoder, some_value);
- *      if (!err)
+ *      if (err)
  *          return err;
  *      err = cbor_encoder_close_container_checked(&encoder, &mapEncoder);
- *      if (!err)
+ *      if (err)
  *          return err;
  *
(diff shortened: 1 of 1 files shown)


src/cborencoder.c
```diff
@@ -117,16 +117,16 @@
  *      CborEncoder encoder, mapEncoder;
  *      cbor_encoder_init(&encoder, buf, sizeof(buf), 0);
  *      err = cbor_encoder_create_map(&encoder, &mapEncoder, 1);
- *      if (!err)
+ *      if (err)
```
georgen117: This will work.  However, the return type is not actually a boolean its an enum.  Shouldn't we be using the enum values?

(i.e. `if (err != CborNoError) return err;` )

src/cborencoder.c
```diff
@@ -157,7 +157,7 @@
  *
  *         cbor_encoder_init(&encoder, buf, size, 0);
  *         err = cbor_encoder_create_array(&encoder, &arrayEncoder, n);
- *         cbor_assert(err);         // can't fail, the buffer is always big enough
+ *         cbor_assert(!err);         // can't fail, the buffer is always big enough
```
georgen117: same comment here shouldn't this be `cbor_assert(CborNoError);`?

src/cborencoder.c
thiagomacieira: `CborNoError` is specifically the value 0 so this kind of construct is possible. But if someone contributes a cleanup to compare to `CborNoError`, I'll probably accept.

thiagomacieira:
The change is good, but I need you to recreate the commit and add a Signed-off-by footer. Can you update?

elie-elkhoury:
> The change is good, but I need you to recreate the commit and add a Signed-off-by footer. Can you update?

I've amended the commit.

thiagomacieira:
Travis seems to be stuck. Bypassing for a simple doc update.

PR #171, opened 2019-11-13
Allow to override install prefix in makefile

deiv:
This way, the makefile can install in the system directories (eg: distribution packages)


thiagomacieira:
The prefix is overridable.
```sh
make prefix=/usr install
```

deiv:
Sorry, my bad :)

PR #174, opened 2020-01-23
Proper #if to remove compilation warning when -Wundef is set

limpkin:
... and remove a const warning

### DIFF
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -44,7 +44,7 @@
 #  include <stdbool.h>
 #endif
 
-#if __STDC_VERSION__ >= 201112L || (defined(__cplusplus) && __cplusplus >= 201103L) || __cpp_static_assert >= 200410
+#if __STDC_VERSION__ >= 201112L || (defined(__cplusplus) && __cplusplus >= 201103L) || (defined(__cpp_static_assert) && __cpp_static_assert >= 200410)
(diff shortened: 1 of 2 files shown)


src/compilersupport_p.h
```diff
@@ -44,7 +44,7 @@
 #  include <stdbool.h>
 #endif
 
-#if __STDC_VERSION__ >= 201112L || __cplusplus >= 201103L || __cpp_static_assert >= 200410
+#if __STDC_VERSION__ >= 201112L || (defined(__cplusplus) && __cplusplus >= 201103L) || __cpp_static_assert >= 200410
```
erichkeane: You'll likely need a similar change to the 3rd component here, __cpp_static_assert is pretty old, but still should probably be guarded in the same way (particularly since it wouldn't be included if someone compiles with -std=c++98).

limpkin: done!

thiagomacieira:
The change looks good, but can you write a little more in your commit messages? Not just "Update <filename>". For simple warning fixes, I usually write:

```
Fix warning: paraphrase warning reason (like your PR title)

filename.c.: warning: full text of the message

Signed-off-by: Name <email@example.com>
```

(Full name not required, just use what you use as the Git commit author)

limpkin:
will do!

PR #175, opened 2020-01-30
Fix compilation warnings

limpkin:
cbortojson.c
Fix warning: initialization discards 'const' qualifier from pointer target type

compilersupport_p.h 
Fix warning: __cplusplus / __cpp_static_assert is not defined …


thiagomacieira:
Bypassing Travis. Looks like Travis is not even scheduling our jobs anymore. I guess I'll need to learn GitHub Actions and replace Travis...

PR #179, opened 2020-02-23
Warning removals

limpkin:
- warning removal: redundant redecleration of open_memstream (in stdio)
- warning removal: function might be possible candidate for gnu_printf
- warning removal: cast from void* to uintptr_t (can't be guaranteed without loss of information in c99

from: Mathieu / mathieu[at]themooltipass[dot]com

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -1093,7 +1093,8 @@ static uintptr_t iterate_memcmp(char *s1, const uint8_t *s2, size_t len)
 
 static uintptr_t iterate_memcpy(char *dest, const uint8_t *src, size_t len)
 {
-    return (uintptr_t)memcpy(dest, src, len);
+    memcpy(dest, src, len);
+    return (uintptr_t)dest;
 }
 
 static CborError iterate_string_chunks(const CborValue *value, char *buffer, size_t *buflen,
--- a/src/cborpretty_stdio.c
+++ b/src/cborpretty_stdio.c
@@ -26,6 +26,7 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+static CborError cbor_fprintf(void *out, const char *fmt, ...) __attribute__ ((format (gnu_printf, 2, 3)));
 static CborError cbor_fprintf(void *out, const char *fmt, ...)
 {
     int n;
--- a/src/cborpretty_stdio.c
+++ b/src/cborpretty_stdio.c
@@ -26,7 +26,9 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#ifdef __GNUC__
 static CborError cbor_fprintf(void *out, const char *fmt, ...) __attribute__ ((format (gnu_printf, 2, 3)));
+#endif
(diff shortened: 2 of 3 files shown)


src/cborparser.c
```diff
@@ -1093,7 +1093,8 @@ static uintptr_t iterate_memcmp(char *s1, const uint8_t *s2, size_t len)
 
 static uintptr_t iterate_memcpy(char *dest, const uint8_t *src, size_t len)
 {
-    return (uintptr_t)memcpy(dest, src, len);
+    memcpy(dest, src, len);
+    return (uintptr_t)dest;
```
thiagomacieira: I don't want this change. It's supposed to be the same thing, but this causes the `iterate_memcpy` function to require a stack frame, save the `dest` pointer somewhere, perform the call and then restore the value. The current implementation is a simple tail-call jump to `memcpy`.

What does you compiler mean with "can't be guaranteed without loss of information"? The definition of `uintptr_t`  says (C11 7.20.1.4):
> The following type designates an unsigned integer type with the property that any valid pointer to **void** can be converted to this type, then converted back to pointer to **void**, and the result will compare equal to the original pointer:

What compiler and architecture was this emitted for?

src/cborpretty_stdio.c
```diff
@@ -26,7 +26,9 @@
 #include <stdarg.h>
 #include <stdio.h>
 
+#ifdef __GNUC__
```
thiagomacieira: Please squash this into the commit that added the `__attribute__`

src/cbortojson.c
```diff
@@ -146,8 +146,6 @@
  * the keys for the metadata clash with existing keys in the JSON map.
  */
 
-extern FILE *open_memstream(char **bufptr, size_t *sizeptr);
```
thiagomacieira: This will break the build in systems where `open_memstream` isn't defined, that is, where we add src/open_memstream.c to the build. For example, like on macOS, which is part of our CI.

We need a new a way to inject this declaration into the build. The easiest is probably a -D flag next to this part of Makefile:
```make
ifeq ($(open_memstream-pass),)
  ifeq ($(funopen-pass)$(fopencookie-pass),)
    CFLAGS += -DWITHOUT_OPEN_MEMSTREAM
    ifeq ($(wildcard .config),.config)
        $(warning warning: funopen and fopencookie unavailable, open_memstream can not be implemented and conversion to JSON will not work prope
rly!)
    endif
  else
    TINYCBOR_SOURCES += src/open_memstream.c
  endif
endif
endif
```

thiagomacieira:
BTW, thanks for the submissions. It's appreciated that people are trying to improve the quality of TinyCBOR!

src/cborparser.c
limpkin: arm eabi (gcc)
you're indeed right... that's a tricky one

thiagomacieira:
Cherry-picking the attribute part. The rest I don't want.

PR #180, opened 2020-03-27
Refine handling of container state synchronisation

sgebbie:
When a container is closed, the outer container's buffer state is
synchronised with the inner container's state.

This was previously done via:

```
CborError cbor_encoder_close_container(CborEncoder *encoder, const CborEncoder *containerEncoder)
{
    if (encoder->end)
        encoder->data.ptr = containerEncoder->data.ptr;
    else
        encoder->data.bytes_needed = containerEncoder->data.bytes_needed;
    encoder->end = containerEncoder->end;

    ...

    return CborNoError;
}
```

However, strictly speaking the inner container could have updated `end` to be `NULL` if
the buffer was too small. In that case, the outer container should carry over `bytes_needed`
and not `ptr`. However, the logic was using the outer container's "stale" view of `end`.

However, generally `sizeof(ptr)` and `sizeof(bytes_needed)` are the same, and these values
exist in a union. Therefore, the correct values where still being copied.

This change moves the synchronisation of `end` to ahead of the synchronisation of `data`.

It additionally, actually avoids this potential error by simply synchronising the full `data`
structure, thereby avoiding the conditional logic. Simplifying the code to simply:

```
encoder->end = containerEncoder->end;
encoder->data = containerEncoder->data;
```

### DIFF
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -540,11 +540,10 @@ CborError cbor_encoder_create_map(CborEncoder *encoder, CborEncoder *mapEncoder,
  */
 CborError cbor_encoder_close_container(CborEncoder *encoder, const CborEncoder *containerEncoder)
 {
-    if (encoder->end)
-        encoder->data.ptr = containerEncoder->data.ptr;
-    else
-        encoder->data.bytes_needed = containerEncoder->data.bytes_needed;
+    // synchronise buffer state with that of the container
     encoder->end = containerEncoder->end;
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Uh, just noticed this MR open. I was on vacation when it came and I must have missed the email... sorry about that! Feel free to post a "ping" whenever a maintainer isn't responsive within reasonable time, myself included.

sgebbie:
Hi Thiago, perfect, thanks for the review and updates.

PR #184, opened 2020-05-30
Enable build for different c libraries

ricardocrudo:
The open_memstream.c is using GLIBC macro definition to test if the library is building on a Linux box. This makes it impossible to build tinycbor against other C libraries, as musl for example. This can be tested using an alpine docker image.


thiagomacieira:
Checked that Bionic and uClibc also have `open_memstream`, so approving.

PR #186, opened 2020-07-27
Fix -Werror,-Wself-assign on big-endian systems

arichardson:
I was getting the following error when compiling cborencoder.c as part of
QtBase for FreeBSD MIPS64:
error: explicitly assigning value of variable of type 'uint64_t' (aka 'unsigned long') to itself [-Werror,-Wself-assign]
    v = cbor_htonll(v);

### DIFF
--- a/src/cborencoder.c
+++ b/src/cborencoder.c
@@ -210,8 +210,8 @@ void cbor_encoder_init(CborEncoder *encoder, uint8_t *buffer, size_t size, int f
 
 static inline void put16(void *where, uint16_t v)
 {
-    v = cbor_htons(v);
-    memcpy(where, &v, sizeof(v));
+    uint16_t v_be = cbor_htons(v);
+    memcpy(where, &v_be, sizeof(v_be));
 }
 
 /* Note: Since this is currently only used in situations where OOM is the only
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Change is correct. I need to see why AppVeyor is failing.

thiagomacieira:
I can reproduce the issues here.

src/cborencoder.c
```diff
@@ -227,14 +227,14 @@ static inline bool isOomError(CborError err)
...
-    memcpy(where, &v, sizeof(v));
+    uint32_t v_be = cbor_htonl(v);
+    memcpy(where, &v_be, sizeof(v_be));
 }
 
 static inline void put64(void *where, uint64_t v)
 {
-    v = cbor_htonll(v);
-    memcpy(where, &v, sizeof(v));
+    uint64_t v_be = cbor_htonll(v_be);
```
thiagomacieira, line 236: Should be `cbor_htonll(v)`

arichardson: Ah yes, sorry about that. Had the right patch in QtBase, but I didn't apply it correctly using the github web ui editor.

thiagomacieira:
Thanks. AppVeyor passed. Let's give Travis an hour.... I don't think Travis works anymore.

thiagomacieira:
I'll merge into dev and update qtbase after this.

arichardson:
> I'll merge into dev and update qtbase after this.

Thanks! This is not that urgent since there are still a few more QtBase patches I need to submit first before MIPS64 FreeBSD builds will work.

thiagomacieira:
Well, Travis certainly took its sweet time, but it checked your change. Thanks, merging.

Do you think you'll have more for TinyCBOR?

arichardson:
> Well, Travis certainly took its sweet time, but it checked your change. Thanks, merging.
> 
> Do you think you'll have more for TinyCBOR?

I don't think there will be any more changes for FreeBSD/MIPS.

However, I may submit more changes once I start running tests for CHERI on MIPS&RISC-V (extending the ISA with 128-bit unforgeable bounds checked pointers, http://www.chericpu.com).

PR #187, opened 2020-08-10
Introduce API for outputting JSON to stream

soburi:
Add  cbor_value_to_json_stream() function to output to a stream for JSON.
It has similar behavior with  cbor_value_to_pretty_stream(),
output JSON text instead of pretty formatted text.

cbor_value_to_pretty_stream() that is used to mapkey-to-text convertion
can take over output stream continuously from JSON output process.
Converted text can output directly to a stream,
no more need to buffering with memstream.

The modification makes the source more portable
and reduces the frequents of dynamic memory allocations.


thiagomacieira:
3 days and Travis hasn't run yet...

soburi:
Travis is back!
I hope you can review at it when you have time.

thiagomacieira:
Will do. Just really busy at work.

PR #191, opened 2020-10-21
Fix links to shared library upon "make install"

h3xx:
GNU install makes a file at the destination if you set the source to a link. This fixes multiply-installed libraries in `/usr/lib` when running `make install`.


thiagomacieira (approved):
Looks good. Travis is taking a long time, so I will wait only 48 hours for it to approve. I don't think we test installation anyway.

h3xx:
I forgot something - test whether installing shared libs. It'll make links even if shared libs are off.

PR #193, opened 2020-12-22
cborinternal_p.h: Fix AVX2 build with MSVC

### LINKED ISSUE #192: Failure to compile with MSVC with /arch:AVX2
```
C:\data\tmp\qt\qtbase\src\3rdparty\tinycbor\src\cborinternal_p.h(44): error C3861: '_cvtss_sh': identifier not found
C:\data\tmp\qt\qtbase\src\3rdparty\tinycbor\src\cborinternal_p.h(48): error C3861: '_cvtsh_ss': identifier not found
```

From: https://bugreports.qt.io/browse/QTBUG-89650

thiagomacieira:
MSVC (and I think ICC too) are lacking the simpler, scalar instructions
to convert from single-precision to half-precision and back. Instead, we
need to use the packed data intrinsics.

Fixes #192.


thiagomacieira:
self-approving

PR #197, opened 2021-01-25
Travis docs fix #196

### LINKED ISSUE #196: travis.yml for MacOS tried to update documentation
The travis.yml executes ./scripts/update-docs.sh.
I think that this should have a guard such that it runs only when within the primary tinycbor repository and/or it should be followed by || exit 0, as I don't think failure to update the docs should be a cause for the regression tests to fail.

It also seems to fail for macOS build only.  Maybe because the listed code sizes changes on macOS only, so that's the only case where the docs need to be updated.

I can send a PR, if you advise how you think you want this fixed.

mcr:
Only build docs if branch is "master"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,11 @@
+env:
+  - BUILD_DOCS=false
+jobs:
+  include:
+    - # only build docs on master
+      if: branch = master
+      env: BUILD_DOCS=true
+
 language: cpp
 matrix:
   include:
@@ -81,4 +89,5 @@ script:
         (cd tests && make TESTARGS=-silent check -k
         TESTRUNNER=`which valgrind 2>/dev/null`)
   - make -s clean
-  - ./scripts/update-docs.sh
+  - [ $BUILD_DOCS ] && ./scripts/update-docs.sh
+


.travis.yml
```diff
@@ -1,3 +1,11 @@
+env:
+  - BUILD_DOCS=false
+jobs:
+  include:
+    - # only build docs on master
+      if: branch = master
```
thiagomacieira, line 6: "master" or "dev" please.

.travis.yml
```diff
@@ -81,4 +89,5 @@ script:
         (cd tests && make TESTARGS=-silent check -k
         TESTRUNNER=`which valgrind 2>/dev/null`)
   - make -s clean
-  - ./scripts/update-docs.sh
+  - [ $BUILD_DOCS ] && ./scripts/update-docs.sh
```
thiagomacieira, line 92: We'll see what Travis says, but if `$BUILD_DOCS` is false, this entire line is false and thus could be considered a CI failure.

thiagomacieira (changes requested):

.travis.yml
mcr: Oh, yeah. Damn.  I put the || exit 0 there before.  I guess this maybe has to be if $BUILD_DOCS; then ... ; fi

thiagomacieira: Or invert:
```yaml
 - [ $SKIP_DOCS ] || ./scripts/update-docs.sh
```

PR #199, opened 2021-01-25
it is unclear what the return value of iterate_memcpy should be.

mcr:
Probably just true, but without this additional cast, I get a warning.

### DIFF
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -1093,7 +1093,7 @@ static uintptr_t iterate_memcmp(char *s1, const uint8_t *s2, size_t len)
 
 static uintptr_t iterate_memcpy(char *dest, const uint8_t *src, size_t len)
 {
-    return (uintptr_t)memcpy(dest, src, len);
+    return (uintptr_t)(char *)memcpy(dest, src, len);
 }
 
 static CborError iterate_string_chunks(const CborValue *value, char *buffer, size_t *buflen,


thiagomacieira (commented):
`memcpy` returns `(void *)` that is the same as the first argument (the destination).

`iterate_string_chunks` will call it expecting a boolean value, like so:
```c
            *result = !!func(buffer + total, (const uint8_t *)ptr, chunkLen);
```

Since `memcpy` always returns non-null, this is a constant true. That's just a trick so I could write smaller code and reuse the same infra for copying and for comparing.

The intermediary cast you're adding here doesn't do anything.

thiagomacieira:
If you want to change this to add a comment explaining what I just said, I'll aprove. The intermediary cast isn't useful.

mcr:
> If you want to change this to add a comment explaining what I just said, I'll aprove. The intermediary cast isn't useful.

It falls -Wall otherwise.
I think that comparing memcpy() results != NULL, would also work, and the compiler would know how to optimize it.

thiagomacieira:
> > If you want to change this to add a comment explaining what I just said, I'll aprove. The intermediary cast isn't useful.
> 
> It falls -Wall otherwise.
> I think that comparing memcpy() results != NULL, would also work, and the compiler would know how to optimize it.

If you're fixing a warning, include the warning text it the commit message.

I didn't want to do the comparison against NULL inside the function because then the compiler couldn't do a tail-call optimisation.

thiagomacieira:
There doesn't seem to be a warning being fixed here.

PR #200, opened 2021-01-25
clarify that CborIndefiniteLength creates an indefinite map.

mcr:
Another interpretation is that the library will guess the number of items and back-patch the length.
(It won't, but it's good to say so)


thiagomacieira (changes requested):
Can you apply the same wording to `create_array` just above?

PR #201, opened 2021-02-09
add CMake support

danewalton:
Added some basic CMake support to the library to make it easier to integrate with other projects. Also added a few bits to README's so let me know if those tweaks are out of scope.


backup-isaac:
Since `open_memstream` doesn't have a Windows implementation, CMake shouldn't try to build it on Windows. I am not the most proficient at CMake so there might be a more idiomatic way to do this, but something like the following should match the behavior of the original Makefiles:

```cmake
set(tinycbor_SOURCES)
list(
    APPEND tinycbor_SOURCES
    ${CMAKE_CURRENT_LIST_DIR}/src/cborencoder.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborencoder_close_container_checked.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborerrorstrings.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborparser.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborparser_dup_string.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborpretty.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborpretty_stdio.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cbortojson.c
    ${CMAKE_CURRENT_LIST_DIR}/src/cborvalidation.c
)

if("${CMAKE_SYSTEM_NAME}" STREQUAL "Windows")
    add_definitions(-DWITHOUT_OPEN_MEMSTREAM)
    message(
        WARNING
        "funopen and fopencookie unavailable, open_memstream can not be implemented and conversion to JSON will not work properly!"
    )
else()
    list(APPEND tinycbor_SOURCES ${CMAKE_CURRENT_LIST_DIR}/src/open_memstream.c)
endif()
```

backup-isaac:
> Since `open_memstream` doesn't have a Windows implementation ...

I made this change in another fork: https://github.com/backup-isaac/tinycbor/tree/cmake-windows

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
+# /****************************************************************************
+# **
+# ** Copyright (C) 2015 Intel Corporation
```
thiagomacieira, line 3: This file is not copyright Intel, since you've written it. Please update with your copyright.

CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+# **
+# ****************************************************************************/
+
+project(intel_tinycbor LANGUAGES C)
+
+cmake_minimum_required(VERSION 3.10)
+
+option(build_tools "Build the cbor tools. Note: This will install the cjson library as a dependency" OFF)
+
+add_library(tinycbor
```
thiagomacieira, line 31: File listing has changed.

examples/testcbor
```diff
@@ -0,0 +1 @@
+��
```
thiagomacieira, line 1: This file shouldn't be here.

thiagomacieira (changes requested):

CMakeLists.txt
danewalton: I think all the `.c` files should be there.

CMakeLists.txt
danewalton: Are you sure? Not sure if you want Microsoft copyright in the intel repo? Since this is a contribution I figured once it's in this repo it is under Intel's control and ownership.

danewalton:
closing in favor of #209 to point to `main`

CMakeLists.txt
thiagomacieira: I sure do. You're creating it and there's no copyright assignment. Therefore, it belongs to you or (in case of work-for-hire, which is your case) the company that hired you.

CMakeLists.txt
thiagomacieira: There are new files in the 0.6 release. That's what needs updating here.

PR #204, opened 2021-03-05
Support arduino api

### LINKED ISSUE #1: Parser Unable to leave a container without going through everything
I'm attempting to write a parser, and it seems that the cbor_value_leave_container is only allowed when you're at the end of the container.  I would prefer to be able to leave the container at any time, so that I only need to pick up the necessary data in an array or map.

soburi:
Support new ArduinoAPI.h that defines arduino namespace.
Fix #1


soburi:
Mistakes to select pushing repository. takedown.

PR #205, opened 2021-03-16
add cbor_value_ref_{text,byte}_string

petm-dec:
add functions to directly obtain a pointer to the data
of a text or byte string, without the need to copy.

otherwise these are analog to the corresponding _copy_ functions.


thiagomacieira:
Zero-copy API is already implemented in https://github.com/intel/tinycbor/compare/dev...thiagomacieira:dev. I've just never published it because it doesn't work really well for microcontrollers.

Since I haven't got the necessary help from them for that in the last 3 years, I'm simply going to ignore them and publish the API as-is.

kalcutter:
Is there some way to help getting the new API merged?... What specifically doesn't work well for microcontrollers?

thiagomacieira:
The problem is that TinyCBOR requires a single buffer with all of your data in it. On many small OSes (I know Zephyr), memory allocation is preferably done in chunks of a fixed size, so if you want for example 200 bytes, you get two 120-byte chunks and form a linked-list with them.

The parsing API was meant to help fix that, but if you couple that with the string API, it gets messy. Suppose you have that 2x120-byte linked list and your string starts 5 bytes into the first one: how does the API return that information?

PR #208, opened 2021-09-04
Document the abstract reader/writer interface

sjlongland:
`tinycbor` fork branch `thiagomacieira/dev` adds support for abstract readers and writers, but the implementation had some limitations in cases where multiple `CborValue` cursors were iterating over the same CBOR document, causing state contamination.  The interface also was not documented.

This documents the new interface and further builds upon it by slightly re-arranging `CborParser` members and opening the `token` in `CborValue` to use by the reader interface for any required purpose.

This pull request is now re-based on `intel/dev` and supercedes https://github.com/thiagomacieira/tinycbor/pull/2.

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -319,8 +319,6 @@ enum CborParserIteratorFlags
 };
 
 struct CborValue;
-
-
 struct CborParserOperations
 {
     bool (*can_read_bytes)(void *token, size_t len);
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -333,79 +333,12 @@ enum CborParserIteratorFlags
 
 struct CborValue;
 
-/**
- * Defines an interface for abstract document readers.  This structure is used
- * in conjunction with \ref cbor_parser_init_reader to define how the various
- * required operations are to be implemented.
- */
+
 struct CborParserOperations
 {
-    /**
-     * Determines whether \a len bytes may be read from the reader.  This is
-     * called before \ref read_bytes and \ref transfer_bytes to ensure it is safe
-     * to read the requested number of bytes from the reader.
-     *
-     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
-     *                  that may be used to pass context information between the
-     *                  \ref CborParserOperations methods.
-     *
-     * \param   len     The number of bytes sought.
-     *
-     * \retval  true    \a len bytes may be read from the reader.
-     * \retval  false   Insufficient data is available to be read at this time.
-     */
     bool (*can_read_bytes)(void *token, size_t len);
-
-    /**
-     * Reads \a len bytes from the reader starting at \a offset bytes from
-     * the current read position and copies them to \a dst.  The read pointer
-     * is *NOT* modified by this operation.
-     *
-     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
-     *                  that may be used to pass context information between the
-     *                  \ref CborParserOperations methods.
-     *
-     * \param   dst     The buffer the read bytes will be copied to.
-     *
-     * \param   offset  The starting position for the read relative to the
-     *                  current read position.
-     *
-     * \param   len     The number of bytes sought.
-     */
     void *(*read_bytes)(void *token, void *dst, size_t offset, size_t len);
-
-    /**
-     * Skips past \a len bytes from the reader without reading them.  The read
-     * pointer is advanced in the process.
-     *
-     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
-     *                  that may be used to pass context information between the
-     *                  \ref CborParserOperations methods.
-     *
-     * \param   len     The number of bytes skipped.
-     */
     void (*advance_bytes)(void *token, size_t len);
-
-    /**
-     * Overwrite the user-supplied pointer \a userptr with the address where the
-     * data indicated by \a offset is located, then advance the read pointer
-     * \a len bytes beyond that point.
-     *
-     * This routine is used for accessing strings embedded in CBOR documents
-     * (both text and binary strings).
-     *
-     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
-     *                  that may be used to pass context information between the
-     *                  \ref CborParserOperations methods.
-     *
-     * \param   userptr The pointer that will be updated to reference the location
-     *                  of the data in the buffer.
-     *
-     * \param   offset  The starting position for the read relative to the
-     *                  current read position.
-     *
-     * \param   len     The number of bytes sought.
-     */
     CborError (*transfer_string)(void *token, const void **userptr, size_t offset, size_t len);
 };
 
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -214,19 +214,6 @@ typedef enum CborEncoderAppendType
     CborEncoderAppendRawData = 2
 } CborEncoderAppendType;
 
-/**
- * Writer interface call-back function.  When there is data to be written to
- * the CBOR document, this routine will be called.  The \a token parameter is
- * taken from the \a token argument provided to \ref cbor_encoder_init_writer
- * and may be used in any way the writer function sees fit.
- *
- * The \a data parameter contains a pointer to the raw bytes to be copied to
- * the output buffer, with \a len specifying how long the payload is, which
- * can be as small as a single byte or an entire (byte or text) string.
- *
- * The \a append parameter informs the writer function whether it is writing
- * a string or general CBOR data.
- */
 typedef CborError (*CborEncoderWriteFunction)(void *token, const void *data, size_t len, CborEncoderAppendType append);
 
 enum CborEncoderFlags
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -319,11 +319,80 @@ enum CborParserIteratorFlags
 };
 
 struct CborValue;
+
+/**
+ * Defines an interface for abstract document readers.  This structure is used
+ * in conjunction with \ref cbor_parser_init_reader to define how the various
+ * required operations are to be implemented.
+ */
 struct CborParserOperations
 {
+    /**
+     * Determines whether \a len bytes may be read from the reader.  This is
+     * called before \ref read_bytes and \ref transfer_bytes to ensure it is safe
+     * to read the requested number of bytes from the reader.
+     *
+     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
+     *                  that may be used to pass context information between the
+     *                  \ref CborParserOperations methods.
+     *
+     * \param   len     The number of bytes sought.
+     *
+     * \retval  true    \a len bytes may be read from the reader.
+     * \retval  false   Insufficient data is available to be read at this time.
+     */
     bool (*can_read_bytes)(void *token, size_t len);
+
+    /**
+     * Reads \a len bytes from the reader starting at \a offset bytes from
+     * the current read position and copies them to \a dst.  The read pointer
+     * is *NOT* modified by this operation.
+     *
+     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
+     *                  that may be used to pass context information between the
+     *                  \ref CborParserOperations methods.
+     *
+     * \param   dst     The buffer the read bytes will be copied to.
+     *
+     * \param   offset  The starting position for the read relative to the
+     *                  current read position.
+     *
+     * \param   len     The number of bytes sought.
+     */
     void *(*read_bytes)(void *token, void *dst, size_t offset, size_t len);
+
+    /**
+     * Skips past \a len bytes from the reader without reading them.  The read
+     * pointer is advanced in the process.
+     *
+     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
+     *                  that may be used to pass context information between the
+     *                  \ref CborParserOperations methods.
+     *
+     * \param   len     The number of bytes skipped.
+     */
     void (*advance_bytes)(void *token, size_t len);
+
+    /**
+     * Overwrite the user-supplied pointer \a userptr with the address where the
+     * data indicated by \a offset is located, then advance the read pointer
+     * \a len bytes beyond that point.
+     *
+     * This routine is used for accessing strings embedded in CBOR documents
+     * (both text and binary strings).
+     *
+     * \param   token   An opaque object passed to \ref cbor_parser_init_reader
+     *                  that may be used to pass context information between the
+     *                  \ref CborParserOperations methods.
+     *
+     * \param   userptr The pointer that will be updated to reference the location
+     *                  of the data in the buffer.
+     *
+     * \param   offset  The starting position for the read relative to the
+     *                  current read position.
+     *
+     * \param   len     The number of bytes sought.
+     */
     CborError (*transfer_string)(void *token, const void **userptr, size_t offset, size_t len);
 };
 
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -214,7 +214,20 @@ typedef enum CborEncoderAppendType
     CborEncoderAppendRawData = 2
 } CborEncoderAppendType;
 
-typedef CborError (*CborEncoderWriteFunction)(void *, const void *, size_t, CborEncoderAppendType);
+/**
+ * Writer interface call-back function.  When there is data to be written to
+ * the CBOR document, this routine will be called.  The \a token parameter is
+ * taken from the \a token argument provided to \ref cbor_encoder_init_writer
+ * and may be used in any way the writer function sees fit.
+ *
+ * The \a data parameter contains a pointer to the raw bytes to be copied to
+ * the output buffer, with \a len specifying how long the payload is, which
+ * can be as small as a single byte or an entire (byte or text) string.
+ *
+ * The \a append parameter informs the writer function whether it is writing
+ * a string or general CBOR data.
+ */
+typedef CborError (*CborEncoderWriteFunction)(void *token, const void *data, size_t len, CborEncoderAppendType append);
 
 enum CborEncoderFlags
 {
--- a/src/cborparser.c
+++ b/src/cborparser.c
@@ -132,6 +132,83 @@
  * \endif
  */
 
+/**
+ * \struct CborParserOperations
+ *
+ * Defines an interface for abstract document readers.  This structure is used
+ * in conjunction with \ref cbor_parser_init_reader to define how the various
+ * required operations are to be implemented.
+ *
+ *
+ * \var CborParserOperations::can_read_bytes
+ *
+ * Determines whether \a len bytes may be read from the reader.  This is
+ * called before \ref read_bytes and \ref transfer_bytes to ensure it is safe
+ * to read the requested number of bytes from the reader.
+ *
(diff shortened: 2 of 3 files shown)


sjlongland:
Forced commits to:
1. remove stale commit from the old `thiagomacieira/dev`
2. fix authorship of commits so that Github's PGP signature verification is happy.

src/cbor.h
```diff
@@ -212,7 +212,20 @@ typedef enum CborEncoderAppendType
     CborEncoderApendRawData = 2
 } CborEncoderAppendType;
 
-typedef CborError (*CborEncoderWriteFunction)(void *, const void *, size_t, CborEncoderAppendType);
+/**
```
thiagomacieira, line 215: I prefer the documentation in the .c source file so the header is clean and small. Makes it easier to search for and find things in it, at a glance. Repeats.

src/cbor.h
```diff
@@ -390,8 +390,9 @@ struct CborParser
 {
     union {
         const uint8_t *end;
-        const struct CborParserOperations *ops;
-    } source;
+        void *ctx;
```
thiagomacieira, line 406: Should we add a `cbor_parser_get_user_context()` so the user doesn't need to know about the internals of the struct? Then we don't have to document it either.

tests/parser/tst_parser.cpp
```diff
@@ -757,32 +757,32 @@ void tst_Parser::mapsAndArrays()
...
-    QByteArray data;
-    int consumed;
-};
-
 static const CborParserOperations byteArrayOps = {
     /* can_read_bytes = */ [](const CborValue *value, size_t len) {
-        auto input = static_cast<Input *>(value->parser->data.ctx);
-        return input->data.size() - input->consumed >= int(len);
+        auto data = static_cast<QByteArray *>(value->parser->data.ctx);
+        auto consumed = uintptr_t(value->source.token);
```
thiagomacieira, line 763: I'd prefer the `uintptr_t` on the left side, but won't insist on your changing again.

tests/parser/tst_parser.cpp
```diff
@@ -757,32 +757,32 @@ void tst_Parser::mapsAndArrays()
...
-        return memcpy(dst, input->data.constData() + input->consumed + offset, len);
+        auto data = static_cast<QByteArray *>(value->parser->data.ctx);
+        auto consumed = uintptr_t(value->source.token);
+        return memcpy(dst, data->constData() + consumed + offset, len);
     },
     /* advance_bytes = */ [](CborValue *value, size_t len) {
-        auto input = static_cast<Input *>(value->parser->data.ctx);
-        input->consumed += int(len);
+        auto consumed = uintptr_t(value->source.token);
+        consumed += int(len);
```
thiagomacieira, line 773: `int` is now wrong. Repeats.

tests/parser/tst_parser.cpp
```diff
@@ -757,32 +757,32 @@ void tst_Parser::mapsAndArrays()
...
+        value->source.token = (void*)consumed;
```
thiagomacieira, line 774: Please write `reinterpret_cast<void *>(consumed)`

thiagomacieira (changes requested):
Hi Stuart

Thank you for your patience. This is looking really good. I will not insist on coding style changes in the examples (they're your code) nor on the `auto` usage. The `reinterpret_cast` is a different case, please update that and please move the Doxygen docs to the .c sources (See https://www.doxygen.nl/manual/docblocks.html#structuralcommands).

I will accept the change without the getter-setter functions working on tokens and context. We can discuss which ones to have after this is merged. I'd rather have wrapper functions so the users don't need to know about the internals of the object, so we can later change them if necessary, like you're doing now.

Finally, what's WSHUB-458? Sounds like a JIRA task number. Can you add a link to the commit message body to what this is and remove from the first line?

src/cbor.h
sjlongland: Ahh okay, I normally put docs in the headers out of habit because of packaging: when you ship a binary build, usually you ship a library object (`.a` / `.so` / `.dll`) and headers that describe the functions within; such users can't see comments in the C sources.

`tinycbor` being statically linked into the sources though and _not_ shipped separately, is a bit different.  I'll move the documentation comments. :-)

src/cbor.h
sjlongland: Sounds fair, I'll add some "standard" getters/setters for structure hiding purposes.

src/cbor.h
thiagomacieira: Yeah, conflicting requirements. I prefer the comments in the sources because I usually read the .h itself. Of course, a good IDE can just collapse the comments...

In the olden days, that meant longer compile times for bigger source.

sjlongland:
> Thank you for your patience. This is looking really good. I will not insist on coding style changes in the examples (they're your code) nor on the `auto` usage. The `reinterpret_cast` is a different case, please update that and please move the Doxygen docs to the .c sources (See https://www.doxygen.nl/manual/docblocks.html#structuralcommands).

Ahh, the `auto` and `reinterpret_cast` are copied and pasted from the existing test cases.  I normally avoid such things but was trying to match the style of the existing test cases.  I will fix that. :-)

> Finally, what's WSHUB-458? Sounds like a JIRA task number. Can you add a link to the commit message body to what this is and remove from the first line?

Good point, force of habit due to the fact I was doing this at work.  I'll strip that token from the commit messages.

src/cbor.h
thiagomacieira: Like I said, we can do this after we merge your change.

thiagomacieira:
(oops, accidentally edited instead of quoting)

> Ahh, the `auto` and `reinterpret_cast` are copied and pasted from the existing test cases. I normally avoid such things but was trying to match the style of the existing test cases. I will fix that. :-)

I'm using the Qt rules for `auto`: it has to be clear what it is, for someone reviewing on a dumb tool like GitHub. That usually means obvious (like iterators) or the type is visible on the same line.

tests/parser/tst_parser.cpp
sjlongland: Replaced with `uintptr_t(len)`

tests/parser/tst_parser.cpp
sjlongland: Yep, I've replaced the `auto` references.

tests/parser/tst_parser.cpp
sjlongland: Fixed, and also did the same for the cast done in `transfer_string` as well.

sjlongland:
> I'm using the Qt rules for `auto`: it has to be clear what it is, for someone reviewing on a dumb tool like GitHub. That usually means obvious (like iterators) or the type is visible on the same line.

Agreed, code clarity must be paramount.  I do more C than C++ so not used to using C++ features like `auto` and tend to avoid using it for that reason.

thiagomacieira (approved):
Let's merge and let me see about using this in Qt too.

thiagomacieira:
Gah, Travis hasn't run yet... Need to find someone to write GitHub Actions workflow, though likely that'll be me.

sjlongland:
Ahh `travis.org` is no more… I hit this on the weekend with one of my own projects (`aioax25`) and it's a landmine waiting to go blam for some of the @vrtsystems and @widesky projects (`pyat`, `cachefs`, `hszinc`) that are still set up to do CI on Travis-CI.  (And sadly, Github Actions doesn't hold a candle to Travis-CI, but Travis-CI is expensive now.)

thiagomacieira:
> Ahh `travis.org` is no more… I hit this on the weekend with one of my own projects (`aioax25`) and it's a landmine waiting to go blam for some of the @vrtsystems and @widesky projects (`pyat`, `cachefs`, `hszinc`) that are still set up to do CI on Travis-CI. (And sadly, Github Actions doesn't hold a candle to Travis-CI, but Travis-CI is expensive now.)

Too bad. Thanks for the update. Let me do a manual check locally and then force the merging.

sjlongland:
So, last time we looked at this, we got caught out by Travis CI's shutdown.  I've just done a rebase on the current `dev` branch to poke CI again.

sjlongland:
Second rebase to pick up the changes on `main` that weren't merged to `dev`.

sjlongland:
Okay, after picking up the changes in `main`, we have a merge conflict, so final rebase onto `dev` again to resolve the merge conflict.  This should unify the currently diverging `main` and `dev` branches once more.

dura0ok:
Hello, what status of this PR?
why no activity?

sjlongland:
> Hello, what status of this PR? why no activity?

Probably because attention has been elsewhere.  I've been doing a lot in the last 2 years that have kept me away from watching what `tinycbor` is up to.

thiagomacieira:
> > Hello, what status of this PR? why no activity?
> 
> Probably because attention has been elsewhere. I've been doing a lot in the last 2 years that have kept me away from watching what `tinycbor` is up to.

If you enable "Allow edits from maintainers" in this PR, I can try to rebase it.

sjlongland:
I managed to do a rebase (initially off the wrong branch, but then I cherry-picked it to the right one).  It seems to have passed AppVeyor CI tests, I haven't got a build environment set up for it just now.

examples/bufferedreader.c
```diff
@@ -0,0 +1,783 @@
+/* vim: set sw=4 ts=4 et tw=78: */
```
thiagomacieira: Please add a header to each of the files with the licence text. Copy from the existing sources (not the existing example, that was my bad and I'll fix it). The copyright is **yours** not Intel.

src/cbor.h
```diff
@@ -321,11 +321,80 @@ enum CborParserIteratorFlags
 };
 
 struct CborValue;
+
+/**
+ * Defines an interface for abstract document readers.  This structure is used
+ * in conjunction with \ref cbor_parser_init_reader to define how the various
+ * required operations are to be implemented.
+ */
```
thiagomacieira: I prefer to keep the docs in the .c source, so the header is neat and small. But because we're documenting the members, which is annoying to do elsewhere, I'll allow it here.

src/cbor.h
```diff
@@ -321,11 +321,80 @@ enum CborParserIteratorFlags
...
+     *                  \ref CborParserOperations methods.
+     *
+     * \param   len     The number of bytes skipped.
+     */
     void (*advance_bytes)(void *token, size_t len);
+
+    /**
+     * Overwrite the user-supplied pointer \a userptr with the address where the
+     * data indicated by \a offset is located, then advance the read pointer
+     * \a len bytes beyond that point.
```
thiagomacieira: Add that something like:
```
    This function should return \c CborNoError if there \a len bytes available at \a offset.
    It should return \c CborErrorUnexpectedEOF if there is not enough data. Other error
    conditions will be passed back to the user (e.g., \c CborErrorIO).
```

src/cbor.h
```diff
@@ -334,25 +334,21 @@ struct CborParserOperations
...
-     *                  \ref CborParserOperations methods.
+     * \param   value   The CBOR value being parsed.
      *
      * \param   len     The number of bytes sought.
      *
      * \retval  true    \a len bytes may be read from the reader.
      * \retval  false   Insufficient data is available to be read at this time.
      */
-    bool (*can_read_bytes)(void *token, size_t len);
+    bool (*can_read_bytes)(const struct CborValue *value, size_t len);
```
thiagomacieira: I want to discuss this change. What use-cases do you have in mind?

src/cborinternal_p.h
```diff
@@ -205,15 +205,15 @@ static inline bool can_read_bytes(const CborValue *it, size_t n)
 #ifdef CBOR_PARSER_CAN_READ_BYTES_FUNCTION
             return CBOR_PARSER_CAN_READ_BYTES_FUNCTION(it, n);
 #else
-            return it->parser->source.ops->can_read_bytes(it, n);
+            return it->parser->ops->can_read_bytes(it, n);
```
thiagomacieira: Then we could remove the `CborParserFlag_ExternalSource` flag and just use `it->parser->ops` as a check that it is external, no?

src/cborparser.c
```diff
@@ -345,7 +345,7 @@ uint64_t _cbor_value_decode_int64_internal(const CborValue *value)
 CborError cbor_parser_init(const uint8_t *buffer, size_t size, uint32_t flags, CborParser *parser, CborValue *it)
 {
     memset(parser, 0, sizeof(*parser));
-    parser->source.end = buffer + size;
+    parser->data.end = buffer + size;
```
thiagomacieira: Add:
```c
    parser->ops = CBOR_NULLPTR;
```

tests/parser/tst_parser.cpp
```diff
@@ -803,32 +803,32 @@ void tst_Parser::mapsAndArrays()
...
-        return memcpy(dst, input->data.constData() + input->consumed + offset, len);
+        auto data = static_cast<QByteArray *>(value->parser->data.ctx);
+        auto consumed = uintptr_t(value->source.token);
+        return memcpy(dst, data->constData() + consumed + offset, len);
     },
     /* advance_bytes = */ [](CborValue *value, size_t len) {
-        auto input = static_cast<Input *>(value->parser->data.ctx);
-        input->consumed += int(len);
+        auto consumed = uintptr_t(value->source.token);
+        consumed += int(len);
```
thiagomacieira: Nitpick: use `qsizetype` or `ptrdiff_t` here, to avoid truncation to 32 bits. I don't think we have a test that operates on more than 2 GB of RAM, but we could add one in the future. Repeats below.

src/cbor.h
```diff
@@ -216,7 +216,20 @@ typedef enum CborEncoderAppendType
...
+ * taken from the \a token argument provided to \ref cbor_encoder_init_writer
+ * and may be used in any way the writer function sees fit.
+ *
+ * The \a data parameter contains a pointer to the raw bytes to be copied to
+ * the output buffer, with \a len specifying how long the payload is, which
+ * can be as small as a single byte or an entire (byte or text) string.
+ *
+ * The \a append parameter informs the writer function whether it is writing
+ * a string or general CBOR data.
+ */
```
thiagomacieira: This one please move to the .c
```c
/**
 * \typedef CborEncoderWriteFunction
...
```

thiagomacieira (changes requested):
Thank you, I have reviewed it. Some of the comments are stale because GitHub won't allow me to go back and edit/discard some of them (I'll dismiss after posting this). GitHub is not a nice review tool.

I'd like you to split this PR in three:
1. The pure doc changes for the interfaces as they currently exist, directly in the .c files (I've confirmed the `\var` works in Doxygen)
2. The CborParser changes
3. The CborEncoder changes

For the latter two, we need to write a good motivation for doing this.

examples/bufferedreader.c
sjlongland: I was happy enough to contribute the copyright to Intel, but yeah I can put my name there.

src/cbor.h
sjlongland: Good question… I'll have to re-read the context because it was 2021 when I wrote this.

src/cborinternal_p.h
sjlongland: That might be an option, not sure if the check for the external source was existing and I kept it there, or if I added it, but `it->parser->ops` should be `NULL` if it's internal.

src/cborparser.c
sjlongland: Will do, I thought the `memset` would effectively do that, but being explicit never hurts.

sjlongland:
> Thank you, I have reviewed it. Some of the comments are stale because GitHub won't allow me to go back and edit/discard some of them (I'll dismiss after posting this). GitHub is not a nice review tool.
> 
> I'd like you to split this PR in three:
> 
>     1. The pure doc changes for the interfaces as they currently exist, directly in the .c files (I've confirmed the `\var` works in Doxygen)

No problems, well I'll keep this PR for the doc changes, then we'll pick out the other changes into separate PRs.

src/cbor.h
sjlongland: Looks like this was done in commit 70dce85d.

examples/bufferedreader.c
sjlongland: Just did this in a local branch (as well as `bufferedwriter.c`), I'll be pushing this shortly as a separate PR.

src/cbor.h
sjlongland: Okay, found the commit where that change was made.  136321a03b37068e9a852eb937cc8934114f65df

`CborValue` objects represent where a given read is taking place to read *that* specific `CborValue`.  This allows for tracking of positions within files at a per-value point of view.

In the example buffered reader, it allocates a block of memory that acts as a "window" into the full file.  If a read happens that needs data that's within the window, the reader just `memcpy`'s it out.  Otherwise, if it's outside that window, the reader can attempt to slide that window back or forth as required.

The alternative would be for it to _always_ reload that buffer with content around the pointer of interest, which would be quite I/O intensive.  (My use case was reading/writing files out of flash filesystems like SPIFFS and littleFS.)

src/cborparser.c
thiagomacieira: > Will do, I thought the `memset` would effectively do that, but being explicit never hurts.

It does for every architecture I am aware of, but it's not required. A null pointer does not have to be represented in memory by a pattern of zero bits. Since TinyCBOR is meant to be cross-platform to unusual microcontrollers, we should be thorough.

For everyone else, the compiler should detect a dead store and eliminate. See https://gcc.godbolt.org/z/o757Thdnd

src/cborinternal_p.h
thiagomacieira: Don't change this now. We'll discuss the removal later, and possibly apply as a separate PR. My comment was just for us to start thinking.

examples/bufferedreader.c
thiagomacieira: > I was happy enough to contribute the copyright to Intel, but yeah I can put my name there.

That's unfortunately more complicated than it seems. Copyright assignment requires paperwork... ask anyone who has contributed to GNU software, which the FSF requires paperwork for.

sjlongland: * [bufferedreader.c](https://github.com/intel/tinycbor/pull/326/commits/f98f354de2a1ea088410f0ea6bbc50e7101b01b8)
* [bufferedwriter.c](https://github.com/intel/tinycbor/pull/327/commits/34042b9ba98b3b601fe5badd7d4396be222224e8)

src/cbor.h
sjlongland: This code block has moved to https://github.com/intel/tinycbor/pull/326

src/cborinternal_p.h
sjlongland: Change made in `496f0fb` (realised I should probably use `CBOR_NULLPTR`).

src/cborparser.c
sjlongland: Taken care of in `4b152eb`.

tests/parser/tst_parser.cpp
sjlongland: I note at some point it was changed to `uintptr_t`, but I've just adjusted it to `ptrdiff_t` in `0ec01d2`.

PR #209, opened 2021-09-22
Add CMake support

danewalton:


CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborencoder.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborencoder_close_container_checked.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborerrorstrings.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborparser.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborparser_dup_string.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborpretty.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborpretty_stdio.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cbortojson.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/cborvalidation.c
+    ${CMAKE_CURRENT_LIST_DIR}/src/open_memstream.c
```
thiagomacieira, line 41: This one needs to be behind a configure-time check to see of `open_memstream` already exists in the C library (it does on Linux with glibc).

thiagomacieira (changes requested):
Thanks, this looks much better. Just the copyright lines as discussed in #201 and the configure-time check.

You don't need to run anything, just check either that this compiles:
```c
#define _GNU_SOURCE
#include <stdio.h>
int main() { open_memstream(NULL, NULL);
```
or that this links:
```c
extern int open_memstream();
int main() { return open_memstream(); }
```

The latter is what Makefile.configure does.

CMakeLists.txt
danewalton: Should have been taken care of with the latest push. I have the sample working on both Mac and Windows right now.

danewalton: cc @backup-isaac thanks for the suggestion

alisonrwu:
Is there any progress on this? I'm looking to add tinycbor to [vcpkg](https://github.com/microsoft/vcpkg) and I think this is necessary..

thiagomacieira:
> Is there any progress on this? I'm looking to add tinycbor to [vcpkg](https://github.com/microsoft/vcpkg) and I think this is necessary..

I'm not good at CMake, so it's hard for me to judge this and I ended up postponing learning enough to be able to take action. Sorry about that.

I'll try to find time in the next two weeks.

Danielius1922:
I somehow didn't notice this when I look at this repo... anyway I've basically [ported](https://github.com/iotivity/iotivity-lite/pull/367/files) the whole Makefile into CMake. Only thing that is not there is packaging into a tar archive (CMake has CPack module for that, so the implementation is not hard.) Otherwise everything - building static, shared libraries, helper executables, feestanding build, building with cJSON installed on the system, tests and installation.
I can create PR and walk you through it.

thiagomacieira:
Would be nice. Though it appears the main issue is my time availability...

PR #217, opened 2022-04-01
If the buffer pointer is a null pointer then don't apply an offset to it

hamchapman:
This is undefined behaviour as caught by Xcode 13.2.1:

```
/Users/ham/TinyCborObjc/tinycbor/src/cborparser.c:1214:37: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /Users/ham/TinyCborObjc/tinycbor/src/cborparser.c:1214:37 in 

```


thiagomacieira (approved):
I don't like pedantism for pedantism sake... but I'll accept this.

thiagomacieira:
Thank you for the contribution!

PR #229, opened 2023-02-10
Tools: fix JSON misspelling in cbordump

niooss-ledger:
Hello,
I found a small misspelling while reading the code. Feel free to fix it either by merging this Pull Request or by adding a commit directly (if for example you found other misspellings).


thiagomacieira (approved):
Thanks!

PR #231, opened 2023-02-23
Use _Float16 for half conversions if available

kalcutter:
Implement support for half-precision floating-point conversions using `_Float16` introduced in C11 extension ISO/IEC TS 18661-3.

Recent ARM compilers support _Float16.

### DIFF
--- a/src/cborinternal_p.h
+++ b/src/cborinternal_p.h
@@ -25,11 +25,19 @@
 #ifndef CBORINTERNAL_P_H
 #define CBORINTERNAL_P_H
 
+/* Dependent source files (*.c) must define __STDC_WANT_IEC_60559_TYPES_EXT__
+ * before <float.h> is (transitively) first included.
+ */
+#if !defined(__STDC_WANT_IEC_60559_TYPES_EXT__)
+#  error __STDC_WANT_IEC_60559_TYPES_EXT__ not defined
+#endif
+
 #include "compilersupport_p.h"
 
 #ifndef CBOR_NO_FLOATING_POINT
 #  include <float.h>
 #  include <math.h>
+#  include <string.h>
 #else
 #  ifndef CBOR_NO_HALF_FLOAT_TYPE
 #    define CBOR_NO_HALF_FLOAT_TYPE     1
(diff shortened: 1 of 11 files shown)


src/cborinternal_p.h
```diff
@@ -37,7 +44,25 @@
 #endif
 
 #ifndef CBOR_NO_HALF_FLOAT_TYPE
-#  if defined(__F16C__) || defined(__AVX2__)
+/* Check for FLT16_MANT_DIG using integer comparison. Clang headers incorrectly
+ * define this macro unconditionally when __STDC_WANT_IEC_60559_TYPES_EXT__
+ * is defined (regardless of actual support for _Float16).
+ */
+#  if FLT16_MANT_DIG > 0 || __FLT16_MANT_DIG__ > 0
+#    include <string.h>
```
thiagomacieira, line 52: Can you move this `#include` to the top? It won't hurt there.

thiagomacieira (changes requested):
looks good, just the minor suggestion.

> Recent ARM compilers support _Float16.

The reason for that is that x86 finally got it too (GCC 12 added support) and that C++ is getting this in [P1467](https://wg21.link/p1467)

src/cborinternal_p.h
c0nk: Done

kalcutter:
> The reason for that is that x86 finally got it too (GCC 12 added support) and that C++ is getting this in [P1467](https://wg21.link/p1467)

I didn't mean that "recent". Even 3-year clang supports this on ARM: https://godbolt.org/z/o3T5oWv57

kalcutter:
> The reason for that is that x86 finally got it too (GCC 12 added support) and that C++ is getting this in [P1467](https://wg21.link/p1467)

Nice... clang 15 also has x86 support now finally.

thiagomacieira:
> > The reason for that is that x86 finally got it too (GCC 12 added support) and that C++ is getting this in [P1467](https://wg21.link/p1467)
> 
> Nice... clang 15 also has x86 support now finally.

I didn't think `_Float16` worked up until recently; `__fp16` did though.

src/cborinternal_p.h
```diff
@@ -25,19 +25,44 @@
...
+    _Float16 f = (_Float16)x;
+    memcpy(&h, &f, 2);
+    return h;
+}
+
+static inline float decode_half(unsigned short x) {
+    _Float16 f;
+    memcpy(&f, &x, 2);
+    return (float)f;
+}
```
thiagomacieira, line 64: Sorry, nitpick: the position of the `{` is wrong. Please move them to the first column, on a line by their own.

I should have noticed this before.

thiagomacieira (changes requested):

src/cborinternal_p.h
kalcutter: Done!

thiagomacieira (approved):

PR #242, opened 2023-10-05
Add CMake support

sdebionne:
This PR adds [CMake](https://cmake.org/) support to build and install the library and CMake export package.


thiagomacieira:
Thank you for the patience!

PR #244, opened 2023-10-27
Update .appveyor.yml from QT v6.1 --> v6.5

mofosyne:
Looks like AppVeyor dropped support for v6.1 QT as shown in https://www.appveyor.com/docs/windows-images-software/

Update to latest QT version


mofosyne (commented):
single line fix for CICD to work again

thiagomacieira (approved):

PR #247, opened 2023-11-02
CI: add Github CI

pjonsson:


pjonsson:
My plan was to make a PR of the first commit, see what fails, and then add some commits that disables -Werror/some configurations/etc so the CI starts going green. I'm not sure why the jobs don't run automatically.

If this is merged, the jobs will be red for reasons I believe are unrelated to the CI:
- There is no bool type in C90 and some tests are compiled as that
- Incompatible `-Werror=something` flag on macOS
- The linux-gcc-non-math test requires math.h in cborinternal_p.h (ldexp. INFINITY, NAN)
- Implicit function declaration of cbor_value_to_pretty_advance_flags in src/cbortojson.c and tools/cbordump/cbordump.c
- Probably some other warning that I have forgotten

.github/dependabot.yml
```diff
@@ -0,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: github-actions
+  directory: "/"
+  schedule:
+    interval: "daily"
+  target-branch: "main"
```
thiagomacieira, line 7: Does the dependabot even work for C code? TinyCBOR has no bundled dependencies (which is the correct way to develop software) so Dependabot not going to find anything even if it works for C code.

.github/workflows/build.yml
```diff
@@ -0,0 +1,162 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Clone tinycbor
+      uses: actions/checkout@v4
+
+    - name: install Linux software
+      if: matrix.os == 'ubuntu-latest'
+      run: |
+        # Need a recent Valgrind, otherwise debug info cannot be read.
+        sudo snap install valgrind --classic
```
thiagomacieira, line 94: And "ubuntu-latest"'s valgrind package using APT is too old? Ugh.

.github/workflows/build.yml
```diff
@@ -0,0 +1,162 @@
...
+      run: |
+        wget -nv https://github.com/ccache/ccache/releases/download/v4.8.3/ccache-4.8.3-linux-x86_64.tar.xz
+        sudo tar xf ccache-4.8.3-linux-x86_64.tar.xz -C /usr/bin --strip-components=1 --no-same-owner ccache-4.8.3-linux-x86_64/ccache
+        rm -f ccache-*-linux-x86_64.tar.xz
+
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1.2.10
+      with:
+        key: compilation-${{ matrix.os }}-${{ matrix.build_cfg.name }}
+        max-size: 160M
```
thiagomacieira, line 135: TinyCBOR is tiny. How much do we gain by having this cached?

thiagomacieira (commented):
Thank you! This is very much appreciated.

> My plan was to make a PR of the first commit, see what fails, and then add some commits that disables -Werror/some configurations/etc so the CI starts going green. I'm not sure why the jobs don't run automatically.

They probably need to be merged first. But meanwhile we can see the results in your fork: https://github.com/pjonsson/tinycbor/actions

> * There is no bool type in C90 and some tests are compiled as that

That's not showing up in your fork. Do you have any details?

> * Incompatible `-Werror=something` flag on macOS

Ditto.

> * The linux-gcc-non-math test requires math.h in cborinternal_p.h (ldexp. INFINITY, NAN)

That one I can see. That requires a fix in TinyCBOR itself, so we can merge the CI and then fix on top.

> * Implicit function declaration of cbor_value_to_pretty_advance_flags in src/cbortojson.c and tools/cbordump/cbordump.c

Not seeing this either, but probably easy to fix.

> * Probably some other warning that I have forgotten

.github/dependabot.yml
pjonsson: I'm not sure of the complete capabilities of Dependabot, it supports npm, gradle, maven, and a bunch of other ecosystems.

The part configured in this PR is the ecosystem "github-actions", to keep the Github action versions in this CI up to date.

.github/workflows/build.yml
pjonsson: ubuntu-latest is 22.04 LTS, so it's a Valgrind 3.17 in the distribution (or something like that).

pjonsson:
> > My plan was to make a PR of the first commit, see what fails, and then add some commits that disables -Werror/some configurations/etc so the CI starts going green. I'm not sure why the jobs don't run automatically.
> 
> They probably need to be merged first. But meanwhile we can see the results in your fork: https://github.com/pjonsson/tinycbor/actions
> 
> > * There is no bool type in C90 and some tests are compiled as that
> 
> That's not showing up in your fork. Do you have any details?

Depending on which job you are looking at, it's possible you're looking at a branch where -Werror is more or less disabled, bool is typedefed to int when compiling pre-C99 standards, and qmake on macOS is from qt6.

I don't have a language standard handy, but https://en.cppreference.com/w/c/keyword/_Bool gives some indication that it was C99 that introduced the boolean type.

The build error is on macOS, but the job linked below times out because it takes >45 minutes to build qt5 on macos-11 (macos-11 is no longer supported by homebrew, and qt5 gives angry warnings about not being tested on the default SDK in the macos-latest image (which I think is macos-12)). I'm not sure why qmake from qt6 is prohibited in tinyCBOR, but I think it's easier to fix whatever is required to use qmake from qt6, so one gets binary qt packages from homebrew.
 
> > * Incompatible `-Werror=something` flag on macOS
> 
> Ditto.

https://github.com/pjonsson/tinycbor/actions/runs/6739438498/job/18320911265

.github/workflows/build.yml
pjonsson: Last time I looked, the Github CI machines were dual core Xeon with a low clock frequency and slow IO. So the gain is noticeable.

thiagomacieira: Let me see if I am allowed to accept this third-party action. I think "github.com/intel" may have disabled all third-party actions.

thiagomacieira:
> Depending on which job you are looking at, it's possible you're looking at a branch where -Werror is more or less disabled, bool is typedefed to int when compiling pre-C99 standards, and qmake on macOS is from qt6.

I decided that the API would use `bool`, so we basically require C99 as a minimum. I don't care about C89 right now. If anyone wants to submit patches that remove warnings for those, feel free.

> I don't have a language standard handy, but https://en.cppreference.com/w/c/keyword/_Bool gives some indication that it was C99 that introduced the boolean type.

It was.

> The build error is on macOS, but the job linked below times out because it takes >45 minutes to build qt5 on macos-11 (macos-11 is no longer supported by homebrew, and qt5 gives angry warnings about not being tested on the default SDK in the macos-latest image (which I think is macos-12)). I'm not sure why qmake from qt6 is prohibited in tinyCBOR, but I think it's easier to fix whatever is required to use qmake from qt6, so one gets binary qt packages from homebrew.

I can look into that.

pjonsson:
> > Depending on which job you are looking at, it's possible you're looking at a branch where -Werror is more or less disabled, bool is typedefed to int when compiling pre-C99 standards, and qmake on macOS is from qt6.
> 
> I decided that the API would use `bool`, so we basically require C99 as a minimum. I don't care about C89 right now. If anyone wants to submit patches that remove warnings for those, feel free.

Well, I personally don't care about pre-C99, but I have a patch for bool on pre-C99 since the C90-test doesn't pass otherwise.

> > The build error is on macOS, but the job linked below times out because it takes >45 minutes to build qt5 on macos-11 (macos-11 is no longer supported by homebrew, and qt5 gives angry warnings about not being tested on the default SDK in the macos-latest image (which I think is macos-12)). I'm not sure why qmake from qt6 is prohibited in tinyCBOR, but I think it's easier to fix whatever is required to use qmake from qt6, so one gets binary qt packages from homebrew.
> 
> I can look into that.

I have a patch for switching macOS to qmake from QT 6 as well. I don't have a mac though, so I can only see that the tests pass that were present in Travis still pass, everything else might crash and burn when switching to QT 6 for all I know.

I don't consider the mentioned patches as fixing CI issues though.

.github/dependabot.yml
thiagomacieira: Ok. I don't see the point, but it won't hurt.

.github/workflows/build.yml
thiagomacieira: Sorry, I can't accept this action. It needs to be "actions/" only or a few options.

thiagomacieira:
> Well, I personally don't care about pre-C99, but I have a patch for bool on pre-C99 since the C90-test doesn't pass otherwise.

Please submit it. We can discuss whether to accept in the review.
> > > The build error is on macOS, but the job linked below times out because it takes >45 minutes to build qt5 on macos-11 (macos-11 is no longer supported by homebrew, and qt5 gives angry warnings about not being tested on the default SDK in the macos-latest image (which I think is macos-12)). I'm not sure why qmake from qt6 is prohibited in tinyCBOR, but I think it's easier to fix whatever is required to use qmake from qt6, so one gets binary qt packages from homebrew.
> > 
> > 
> > I can look into that.
> 
> I have a patch for switching macOS to qmake from QT 6 as well. I don't have a mac though, so I can only see that the tests pass that were present in Travis still pass, everything else might crash and burn when switching to QT 6 for all I know.

I've fixed the Clang issue. Please rebase to see it.

pjonsson:
The bool fix is in #249.

> I've fixed the Clang issue. Please rebase to see it.

Rebase against what branch?

thiagomacieira:
> Rebase against what branch?

main. The change is ... uh... where did it go? Oops, it failed and I didn't notice.

Done now, as 0d5538277d9975c448bd819572ca961542777aff in main

.github/workflows/build.yml
pjonsson: Removed.

mofosyne:
Oh is this PR already done, as https://github.com/intel/tinycbor/commit/0d5538277d9975c448bd819572ca961542777aff in main ? If so then should this PR be closed then?

pjonsson:
This PR adds Github CI, https://github.com/intel/tinycbor/commit/0d5538277d9975c448bd819572ca961542777aff makes the Makefile work with Clang.

thiagomacieira:
Right. According to the CI logs in your repository, one of the builds is failing.

I'll merge as-is and then we can figure out how to fix it. It's better than no CI.

pjonsson:
The Github macOS CI runners for "macos-latest" have changed CPUs from x86-64 to M1/M2 in the last month(s), so it's possible things related to macOS broke (more) after I made this PR.

thiagomacieira:
I switched to `macos-13` so Valgrind would run... but Valgrind is broken anyway, so I removed that.

PR #248, opened 2023-11-03
Makefile: add ccache support

pjonsson:
Ccache speeds up recompilation by caching
previous compilations and detecting when
the same compilation is done again.


thiagomacieira:
Thanks, but this isn't really necessary. You can get this by using `make CC="ccache gcc"`.

What's more, I don't think this should be automatic.

PR #249, opened 2023-11-26
Add C90 bool support

pjonsson:
The bool datatype was added in C99,
so make a typedef for pre-C99 versions
of C, and define true/false.

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -45,7 +45,17 @@
 #ifdef __cplusplus
 extern "C" {
 #else
-#include <stdbool.h>
+#if defined __STDC_VERSION__ && __STDC_VERSION__ >= 199901L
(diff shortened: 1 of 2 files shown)


src/cbor.h
```diff
@@ -45,7 +45,17 @@
...
+#if defined __STDC_VERSION__ && __STDC_VERSION__ >= 199901L
+#  include <stdbool.h>
+#else
+   typedef int bool;
+#  ifndef true
+#    define true 1
+#  endif
+#  ifndef false
+#    define false 0
+#  endif
```
thiagomacieira, line 57: This is correct, but such not a good idea in a public header.

I think we're just going ban C89 builds.

thiagomacieira:
Thank you for the effort, but no, I think we will reject.

PR #266, opened 2025-02-07
tst_Parser: add some testing rows for floating point data

thiagomacieira:


tests/parser/tst_parser.cpp
```diff
@@ -38,6 +38,10 @@
 #  include <windows.h>
 #endif
 
+#ifndef QCOMPARE_EQ
+// added for Qt 6.4
+#  define QCOMPARE_EQ QCOMPARE
```
dangelog, line 43: Not that it's a big deal (6.4 is EOL), but IIRC this is not strictly equivalent, floating-point types are compared using qFuzzyCompare with QCOMPARE but exactly with QCOMPARE_EQ?

PR #289, opened 2025-03-19
Remove the old Makefile and qmake buildsystem files

thiagomacieira:


terjr:
Why is Makefile.configure and Makefile.nmake left in repo? They are obsolete now, right?

thiagomacieira:
> Why is Makefile.configure and Makefile.nmake left in repo? They are obsolete now, right?

Oops. Hadn't noticed those two. Thank you for pointing out → #316

PR #295, opened 2025-04-13
Fetchcontent frendly CMakeLists.txt

jhlee525:
I changed CMakeLists.txt to enable `FetchContent` this library from the user library.

1. In the case of `cmake_source_dir`, there was an issue where it was overridden with the parent project path when building the library using `FetchContent`, causing it to not work properly. So I changed `cmake_source_dir` to `project_source_dir`.
2.  I remember the header install path is `tinycbor/cbor.h`, but has the installation path changed to `cbor.h`? If the `cbor.h` installation is only performed during the cmake install process, that part should also be fixed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@ cmake_minimum_required(VERSION 3.10)
 project(tinycbor LANGUAGES C CXX VERSION 7.0)
 
 # Set path to additional cmake scripts
-set(CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
+set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
 
 set(TARGETS_EXPORT_NAME "TinyCBOR-targets")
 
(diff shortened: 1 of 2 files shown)


thiagomacieira:
Can you squash the four commits into one and explain in the commit message why those changes are necessary to make TinyCBOR more FetchContent-friendly? The explanation in the PR text is probably sufficient, but no one looks at PR texts after committing, but they do look at commit messages via `git blame`.

jhlee525:
Changed :)

jhlee525:
However still curious about my question in PR text `2.`

CMakeLists.txt
```diff
@@ -142,14 +142,14 @@ endif()
...
-  PUBLIC "$<BUILD_INTERFACE:${CMAKE_BINARY_DIR}>"
+  PUBLIC "$<BUILD_INTERFACE:${PROJECT_BINARY_DIR}>"
   PUBLIC "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>"
 )
 
 install(FILES
-  src/cbor.h
-  ${CMAKE_BINARY_DIR}/tinycbor-version.h
-  ${CMAKE_BINARY_DIR}/tinycbor-export.h
+  ${PROJECT_SOURCE_DIR}/src/cbor.h
```
thiagomacieira, line 150: I'm surprised this one is needed. Shouldn't paths be relative to the current dir anyway?

jhlee525:
I added issue https://github.com/intel/tinycbor/issues/296

If necessary, I can make following PR to resolve this problem

CMakeLists.txt
friendlyanon: https://cmake.org/cmake/help/latest/command/install.html#files

> File names given as relative paths are interpreted with respect to the current source directory.

thiagomacieira:
> I added issue #296
> 
> If necessary, I can make following PR to resolve this problem

Please do, it's easier for me to review someone else's changes and approve than getting my own reviewed.

CMakeLists.txt
thiagomacieira: We'll leave. I just wanted to know if I was missing something.

Thanks.

PR #298, opened 2025-04-24
set public header include dir for fixing header prefix

jhlee525:
In the previous PR(#297), the install directory for cbor.h was changed to tinycbor/cbor.h.

it causes problem when using FetchContent, since it import headers directly from the source code. (It actually placed in `src/tinycbor.h`).

I created a dedicated public header directory `include` by convention, and placed `include/tinycbor/cbor.h`.

Since all source files reference just `cbor.h`, I have to updated all include paths accordingly.

I checked `tests` and `tools` are able to compile, but not tested for all option sets.

Apologies for the big changes, If you prefer not to adopt this change, feel free to revert #297.


thiagomacieira (changes requested):
I don't like not having the header file right next to the sources. Can you keep the cbor.h header exactly where it is, but instead add include/tinycbor/cbor.h
```c
#include "../../src/cbor.h"
```
?

Why does everything else need changing?

PR #303, opened 2025-08-19
cbortojson: [CWE-401] fix memory leak when realloc() is fails

GermanAizek:
@thiagomacieira, I accidentally found it while I was studying the source code.

It is better to close this possible leak before assign CVE id and score, and also protect yourself from possible attack vectors by intruders.

This commit also have identical issue as real CVE-2019-17178 vulnerability in FreeRDP project:

Fix commit - https://github.com/FreeRDP/FreeRDP/commit/9fee4ae076b1ec97b97efb79ece08d1dab4df29a

Issue - https://github.com/FreeRDP/FreeRDP/issues/5645

More info about CWE metrics:

https://cwe.mitre.org/data/definitions/401.html

### DIFF
--- a/src/cbortojson.c
+++ b/src/cbortojson.c
@@ -329,9 +329,10 @@ static CborError escape_text_string(char **str, size_t *alloc, size_t *offsetp, 
         return CborErrorDataTooLarge;
     }
     if (!alloc || needed > *alloc) {
-        buf = cbor_realloc(buf, needed);
-        if (!buf)
+        char *tmp = cbor_realloc(buf, needed);
+        if (!tmp)
             return CborErrorOutOfMemory;
+        buf = tmp;
         if (alloc)
             *alloc = needed;
     }


src/cbortojson.c
```diff
@@ -329,9 +329,10 @@ static CborError escape_text_string(char **str, size_t *alloc, size_t *offsetp,
         return CborErrorDataTooLarge;
     }
     if (!alloc || needed > *alloc) {
-        buf = cbor_realloc(buf, needed);
-        if (!buf)
+        char *tmp = cbor_realloc(buf, needed);
+        if (!tmp)
             return CborErrorOutOfMemory;
+        buf = tmp;
```
thiagomacieira, line 335: Even if there is a problem with freeing, this patch isn't solving it. `buf` is a local variable in this function, so adding yet another temporary is not going to help. This code will compile to exactly the same assembly before and after the change.

thiagomacieira (changes requested):

src/cbortojson.c
GermanAizek: @thiagomacieira, how do you propose a change so that its more secure?

thiagomacieira:
I'm not seeing a problem here. Can you show an example of where there would be an issue?

PR #305, opened 2025-08-20
cbor: reduce CborEncoder cacheline size to 24 bytes for x86-64 arch

GermanAizek:
@thiagomacieira,
## About patch changes

size_t have 8 byte in x64 arch, flags strict 4 byte (int) therefore, C compiler adding +4 bytes padding in memory layout for struct.

Before this patch CborEncoder have size in diff arch (x86 - 24 bytes, x64 - 32 bytes).

Now size is identical, uint64_t is very difficult to fill, so I shortened range using bit fields. This change may be performance-critical for older processors without L3 cache, low ram capacity or embedded systems.

### DIFF
--- a/src/cbor.h
+++ b/src/cbor.h
@@ -230,8 +230,16 @@ struct CborEncoder
         CborEncoderWriteFunction writer;
     } data;
     uint8_t *end;
+#ifdef __i386__
+    uint32_t remaining;
+    int flags;
+#elifdef __x86_64__
+    uint64_t remaining : 58;
+    int flags : 6;
+#else
     size_t remaining;
     int flags;
+#endif
 };
 typedef struct CborEncoder CborEncoder;
 


src/cbor.h
```diff
@@ -230,8 +230,16 @@ struct CborEncoder
         CborEncoderWriteFunction writer;
     } data;
     uint8_t *end;
+#ifdef __i386__
+    uint32_t remaining;
+    int flags;
+#elifdef __x86_64__
```
thiagomacieira, line 236: `#elifdef` is C23. We can't have that.

thiagomacieira (commented):
> size_t have 8 byte in x64 arch, flags strict 4 byte (int) therefore, C compiler adding +4 bytes padding in memory layout for struct.

Correct, but that's not a problem. In fact, a 32-byte struct is easier to allocate in memory so it won't cross cachelines. Not that I expect applications to have more than a handful of `CborEncoder` objects in the first place: the number should be that of the depth of the most nested item.

> Before this patch CborEncoder have size in diff arch (x86 - 24 bytes, x64 - 32 bytes).
>
> Now size is identical,

No, it isn't and your x86 calculation was wrong. The size on all 32-bit architectures was 16 bytes, not 24.

This patch would reduce the size on x86-64, but why should we do that for x86-64 only? Why not for other 64-bit architectures?

> This change may be performance-critical for older processors without L3 cache, low ram capacity or embedded systems.

x86-64 systems without L3 still running? With low RAM that can't deal with a 32-byte struct in modest quantities? Sorry, but I don't buy that.

The original Athlon64 did come without L3, but it had 64 kB of L1d and 512 kB of L2. It may have had on-motherboard L3, whereas today we have on-die L3, and some of those were already common back then.

src/cbor.h
GermanAizek: > x86-64 systems without L3 still running? With low RAM that can't deal with a 32-byte struct in modest quantities? Sorry, but I don't buy that.

@thiagomacieira, I building and create PR with patch from Celeron D220 from 2008 without L3 cache with 4 GB DDR2.
I also have MODERN from AMD A6-9225 without L3 cache from 2018 and Intel Atom Z3735F without L3 cache from 2016.

They are all 64-bit and do not have an L3 cache. They have enough RAM for modern tasks, it's just that structs is growing in size and no one project is optimizing it.

GermanAizek: AMD A6-9225 from 2018, have AVX2 but not have L3 cache. These CPU in my list are unique in their own way, because CPU power consumption.

thiagomacieira:
You're making the runtime slower on x86-64 for a dubious benefit. As I said, the number of `CborEncoder` objects should be fairly small, so a difference of 8 bytes won't really matter.

PR #319, opened 2026-04-02
support picolibc

abobija:
With current `HAVE_OPEN_FUNOPEN` and `HAVE_OPEN_FOPENCOOKIE` defines i can not configure `src/open_memstream.c` to work with picolibc's version of [`funopen`](https://github.com/picolibc/picolibc/blob/5c5b147244adffdf197c05841f5b4ce1d26f92bb/libc/include/stdio.h#L488).

This PR is just POC (i'm sure there is a cleaner way) how to configure it to work.

picolibc is default libc for espressif's [esp-idf](https://github.com/espressif/esp-idf) from version 6 so it's going to be used there a lot so it would be cool if tinycbor could support picolibc somehow as well.

### DIFF
--- a/src/open_memstream.c
+++ b/src/open_memstream.c
@@ -38,7 +38,7 @@
 #if defined(HAVE_OPEN_FUNOPEN)
 typedef int RetType;
 typedef int LenType;
-#elif defined(HAVE_OPEN_FOPENCOOKIE)
+#elif defined(__PICOLIBC__) || defined(HAVE_OPEN_FOPENCOOKIE)
 typedef ssize_t RetType;
 typedef size_t LenType;
 #else
@@ -54,7 +54,11 @@ struct Buffer
     size_t alloc;
 };
 
+#ifdef __PICOLIBC__
+static RetType write_to_buffer(void *cookie, const void *data, LenType len)
+#else
 static RetType write_to_buffer(void *cookie, const char *data, LenType len)
+#endif
 {
     struct Buffer *b = (struct Buffer *)cookie;
     char *ptr = *b->ptr;
@@ -99,7 +103,7 @@ FILE *open_memstream(char **bufptr, size_t *lenptr)
     *bufptr = NULL;
     *lenptr = 0;
 
-#if defined(HAVE_OPEN_FUNOPEN)
+#if defined(__PICOLIBC__) || defined(HAVE_OPEN_FUNOPEN)
     return funopen(b, NULL, write_to_buffer, NULL, close_buffer);
 #elif defined(HAVE_OPEN_FOPENCOOKIE)
     static const cookie_io_functions_t vtable = {


src/open_memstream.c
```diff
@@ -54,7 +54,11 @@ struct Buffer
     size_t alloc;
 };
 
+#ifdef __PICOLIBC__
+static RetType write_to_buffer(void *cookie, const void *data, LenType len)
+#else
 static RetType write_to_buffer(void *cookie, const char *data, LenType len)
+#endif
```
thiagomacieira, line 61: This is the same thing, so you didn't do anything.

thiagomacieira (changes requested):
Thank you for the POC, but we're going to need to work on this. The patch cannot be right.

Please start with what error messages you're getting.

src/open_memstream.c
abobija: Thank you for response.

Error that I'm facing:

```
tinycbor/src/open_memstream.c:103:29: error: passing argument 3 of 'funopen' from incompatible pointer type [-Wincompatible-pointer-types]
  103 |     return funopen(b, NULL, write_to_buffer, NULL, close_buffer);
      |                             ^~~~~~~~~~~~~~~
      |                             |
      |                             RetType (*)(void *, const char *, LenType) {aka int (*)(void *, const char *, int)}

In file included from tinycbor/src/open_memstream.c:31:
picolibc/include/stdio.h:429:28: note: expected '_ssize_t (*)(void *, const void *, size_t)' {aka 'int (*)(void *, const void *, unsigned int)'} but argument is of type 'RetType (*)(void *, const char *, LenType)' {aka 'int (*)(void *, const char *, int)'}
  429 |                 _ssize_t (*writefn)(void *cookie, const void *buf, size_t n),
      |                 ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The goal is to align signature of tinycbor's `write_to_buffer` function with picolibc's `writefn` function signature:

```c
_ssize_t (*writefn)(void *cookie, const void *buf, size_t n)
```

In my POC, for `__PICOLIBC__`, I've set:
- `RetType` to be `ssize_t`
- `LenType` to be `size_t`

With that I'm still getting error of signature mismatch since second parameter (`data`) of tinycbor's function `write_to_buffer` is type of `const char*` but picolibc expects that param to be type of`const void*`. I've fixed that with this:

```diff
+ #ifdef __PICOLIBC__
+ static RetType write_to_buffer(void *cookie, const void *data, LenType len)
+ #else
static RetType write_to_buffer(void *cookie, const char *data, LenType len)
+ #endif
```

^ you can see there that type of second parameter is changed from `const char*` to `const void*` for picolibc.

Maybe we need new `DataType` beside `RetType` and `LenType` to make `open_memstream.c` more scalable?

Hope I've explain it better this time.

thiagomacieira: Ok, I see the problem: for BSD systems (including Apple's macOS), `funopen`'s callback functions take a `const char *` second parameter, but in picolibc, it's `const void *`. I suppose that makes sense, as we could be writing binary data through `fwrite` and such. However, this is very stupid of picolibc to mismatch the API: if you're trying to implement the API that 4.4 BSD introduced, it should match the signature.

To fix this, we need to update the CMakeLists.txt test to do more than `check_symbol_exists`:
https://github.com/intel/tinycbor/blob/6442e749ca811e24afad1551338a45c75e09808f/CMakeLists.txt#L133-L135

We need a compile check that we are matching the arguments correctly, to be able to match the two variants.

Issue #207, opened 2021-06-09
Error during compilation for NodeMCU (12.1E)


sariyamon:
I am working on ArduinoIoTColud project and I'm using the auto-generated script from the ArduinoIotCloud.
The following error occurred during the compilation.

**Code**
#include "arduino_secrets.h"

_/* 
  Sketch generated by the Arduino IoT Cloud Thing 
  https://create.arduino.cc/cloud/things/

  Arduino IoT Cloud Variables description

  The following variables are automatically generated and updated when changes are made to the Thing

  CloudLight switch1;
  CloudLight switch3;
  CloudTemperatureSensor temp;
  CloudLight switch2;

  Variables which are marked as READ/WRITE in the Cloud Thing will also have functions
  which are called when their values are changed from the Dashboard.
  These functions are generated with the Thing and added at the end of this sketch.
*/

#include "thingProperties.h"

void setup() {
  // Initialize serial and wait for port to open:
  Serial.begin(9600);
  // This delay gives the chance to wait for a Serial Monitor without blocking if none is found
  delay(1500); 

  // Defined in thingProperties.h
  initProperties();

  // Connect to Arduino IoT Cloud
  ArduinoCloud.begin(ArduinoIoTPreferredConnection);
  
  /*
     The following function allows you to obtain more information
     related to the state of network and IoT Cloud connection and errors
     the higher number the more granular information you’ll get.
     The default is 0 (only errors).
     Maximum is 4
 */
  setDebugMessageLevel(2);
  ArduinoCloud.printDebugInfo();
}

void loop() {
  ArduinoCloud.update();
  // Your code here 
  
  
}

void onSwitch1Change() {
  // Do something
}

void onSwitch2Change() {
  // Do something
}

void onSwitch3Change() {
  // Do something
}_

**Error Details**

Arduino: 1.8.15 (Windows 10), Board: "NodeMCU 1.0 (ESP-12E Module), 80 MHz, Flash, Disabled (new aborts on oom), Disabled, All SSL ciphers (most compatible), 32KB cache + 32KB IRAM (balanced), Use pgm_read macros for IRAM/PROGMEM, 4MB (FS:2MB OTA:~1019KB), 2, v2 Lower Memory, Disabled, None, Only Sketch, 115200"

C:\Users\xxx\Documents\Arduino\libraries\ArduinoIoTCloud\src\cbor\lib\tinycbor\src\cborpretty.c: In function 'hexDump':

[...]

PR #320, opened 2026-04-13
fix issue with PRIu64 on some compilers, by including sys/types.h early

AuxLoui:
Fix for #207 (PRIu64 problem)

### DIFF
--- a/src/cborpretty.c
+++ b/src/cborpretty.c
@@ -29,6 +29,9 @@
 #include "compilersupport_p.h"
 #include "utf8_p.h"
 
+// Include sys/types.h before inttypes.h to work around issue with
+// certain versions of GCC and newlib which causes omission of PRIx64
+#include <sys/types.h>
 #include <inttypes.h>
 #include <string.h>
 
--- a/src/cbortojson.c
+++ b/src/cbortojson.c
@@ -31,6 +31,9 @@
 #include "cborinternal_p.h"
 #include <memory.h>
 
+// Include sys/types.h before inttypes.h to work around issue with
+// certain versions of GCC and newlib which causes omission of PRIx64
+#include <sys/types.h>
 #include <inttypes.h>
 #include <stdio.h>
 #include <stdlib.h>


src/cborpretty.c
```diff
@@ -29,6 +29,9 @@
 #include "compilersupport_p.h"
 #include "utf8_p.h"
 
+// Include sys/types.h before inttypes.h to work around issue with
+// certain versions of GCC and newlib which causes omission of PRIx64
```
thiagomacieira, line 33: The comment wasn't necessary. But I'll allow it.

However, is this high enough? Imagine something changes in the headers above and breaks? Especially cbor.h, which includes stdint.h.

thiagomacieira (commented):
Feedback on making commits and making PRs (general, not about this one in particular):
* the commit message should explain why you're making a change. If you have an error message, it's often useful to include it.
* you don't need to explain the *what* you're changing, because we can read diffs
* the first line of the commit is the "subject" and should be a summary, which we can see in one-line logs (e.g. https://github.com/intel/tinycbor/commits/main/)
* some projects require that you include a signed-off-by footer. We don't insist for TinyCBOR, but I would prefer it is there.
* each commit should be a single, logical/semantic change, which doesn't break the build
* the GitHub PR page is a summary of all commits in the change. If there's only one commit, then just copy the commit message and body (GitHub will do it for you).

src/cborpretty.c
AuxLoui: Sorry for late reply. Your comment did have me give an other thought about this problem, and came to the conclusion that instead of implementing a work around for a specific compiler instead you should use a working compiler. So instead of using this fix/workaround I changed my docker image used to compile my program in CI/CD pipeline.  So will close this pull request, if other have this problem this is my used dockerbuild definition used wit gitlab.

```
FROM ubuntu:26.04

ENV DEBIAN_FRONTEND=noninteractive

ARG ARM_GNU_TOOLCHAIN_VERSION=15.2.Rel1
ARG ARM_GNU_TOOLCHAIN_ARCHIVE=arm-gnu-toolchain-15.2.Rel1-x86_64-arm-none-eabi.tar.xz
ARG ARM_GNU_TOOLCHAIN_DIR=arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi
ARG ARM_GNU_TOOLCHAIN_URL=https://developer.arm.com/-/media/Files/downloads/gnu/${ARM_GNU_TOOLCHAIN_VERSION}/binrel/${ARM_GNU_TOOLCHAIN_ARCHIVE}
ARG PICO_SDK_VERSION=2.2.0

RUN apt-get update && apt-get install -y --no-install-recommends \
    build-essential \
    ca-certificates \
    cmake \
    curl \
    git \
    ninja-build \
    python3 \
    xz-utils && \
    rm -rf /var/lib/apt/lists/*

RUN mkdir -p /opt/toolchains /opt/pico-sdk && \
    curl -fsSL "${ARM_GNU_TOOLCHAIN_URL}" -o "/tmp/${ARM_GNU_TOOLCHAIN_ARCHIVE}" && \
    tar -xJf "/tmp/${ARM_GNU_TOOLCHAIN_ARCHIVE}" -C /opt/toolchains && \
    rm -f "/tmp/${ARM_GNU_TOOLCHAIN_ARCHIVE}" && \
    test -x "/opt/toolchains/${ARM_GNU_TOOLCHAIN_DIR}/bin/arm-none-eabi-gcc" && \
    git clone --branch "${PICO_SDK_VERSION}" --depth 1 https://github.com/raspberrypi/pico-sdk.git /opt/pico-sdk && \
    git -C /opt/pico-sdk submodule update --init --recursive

ENV ARM_GNU_TOOLCHAIN_VERSION=${ARM_GNU_TOOLCHAIN_VERSION}
ENV ARM_GNU_TOOLCHAIN_PATH=/opt/toolchains/${ARM_GNU_TOOLCHAIN_DIR}
ENV PICO_SDK_VERSION=${PICO_SDK_VERSION}
ENV PICO_SDK_PATH=/opt/pico-sdk
ENV PATH=${ARM_GNU_TOOLCHAIN_PATH}/bin:${PATH}
```

PR #321, opened 2026-04-13
install json2cbor

florolf:
The old Make-based build system installed json2cbor during "make install" if it has been built. However, this behavior got dropped during the CMake migration, causing this tool to be missing by default. Restore the old behavior.


thiagomacieira (approved):
Thank you.

PR #322, opened 2026-04-13
compilersupport: fix build with GCC < 11

florolf:
Commit 45e4641 ("Fix build with GCC < 11: [[fallthrough]] is supported but not allowed in C") introduced a fix for old GCC versions, but commit 91d1c50 ("compilersupport: fix compilation in C23 mode") reverted it, reintroducing the build failure.

Revert the revert.

This builds fine with both an old toolchain and a new one when I force C23 (despite the top-level CMakeFile normally forcing it to C99). I'm not sure if this was an intentional change or just an accident?

### DIFF
--- a/src/compilersupport_p.h
+++ b/src/compilersupport_p.h
@@ -52,8 +52,12 @@
 #  define cbor_static_assert(x)         ((void)sizeof(char[2*!!(x) - 1]))
 #endif
 
-#if defined(__has_cpp_attribute)    // C23 and C++17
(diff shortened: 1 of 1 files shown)


thiagomacieira (approved):

thiagomacieira:
> I'm not sure if this was an intentional change or just an accident?

Pretty sure it was an accident. There was a point in time when my copy on disk, shared with Qt, went out of sync with upstream TinyCBOR (both dev and main branches). That's also the reason I decided to finally bite the bullet and release (0.)7.0.

PR #323, opened 2026-04-13
CMake: don't depend on C++ compiler for main build

florolf:
Only the tests actually require a C++ compiler, however CXX is globally enabled for the project, which prevents building on targets that do not have a C++ compiler available.

Only enable C++ support selectively when building the tests is enabled too.

Note that I couldn't test this with MSVC, but the conditions should still work if I'm reading the CMake docs correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@
 
 cmake_minimum_required(VERSION 3.10)
 
-project(tinycbor LANGUAGES C CXX VERSION 7.0)
+project(tinycbor LANGUAGES C VERSION 7.0)
 
 # Set path to additional cmake scripts
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake ${CMAKE_MODULE_PATH})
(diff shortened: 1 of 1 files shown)


thiagomacieira (commented):
> Note that I couldn't test this with MSVC, but the conditions should still work if I'm reading the CMake docs correctly.

There is no release of MSVC lacking a C++ compiler. At least as far back as I remember (Microsoft C/C++ 7.0, which was the first release of C++), it's always installed both. So if this continues to compile (AppVeyor CI), it will probably be fine.

thiagomacieira (approved):

thiagomacieira:
We can see in the "gcc-small" CI step that it did not search for the C++ compiler. So this is working.

We could add a CI step to make sure we don't break it, but...

florolf:
For what it's worth, the Buildroot [autobuild](https://autobuild.buildroot.org/) jobs are testing this case, which is how I found the issue in the first place. If it breaks again, somebody over there will probably notice :)

florolf:
Hmm, it seems the automerge didn't trigger (possibly because I added a comment?) could you maybe poke it again @thiagomacieira? Thanks!

thiagomacieira:
> Hmm, it seems the automerge didn't trigger (possibly because I added a comment?) could you maybe poke it again @thiagomacieira? Thanks!

Done and merged.

PR #325, opened 2026-04-23
CMake: fixed `tinycbor.pc` install

EduMenges:
The line `configure_file(tinycbor.pc.in tinycbor.pc @ONLY)` generated the file into `CMAKE_CURRENT_BINARY_DIR`, but the installation of this file was being done using `CMAKE_BINARY_DIR`. This creates an error if tinycbor is added through `add_subdirectory`, because it tries looking that the root binary directory, but doesn't find it there because tinycbor gets its own subdirectory. This should fix the installation by using that specific directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ install(FILES
   DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/tinycbor
 )
 install(FILES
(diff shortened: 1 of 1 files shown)


thiagomacieira (changes requested):
Looks good, but can you put the explanation in the commit message? What you used on the PR looks good.

PR #326, opened 2026-04-25
Implement per-value reader context in abstract reader interface

sjlongland:
The abstract reader interface relies on passing around a `token` which represents the abstract reader context, keeping context information global to the entire parser instance.

This leads to significantly more work at the implementation end as it has no concept of what is being read and therefore whether it might already have the information needed.  The end result being additional I/O that could have been avoided.

By passing in the `CborValue`, we can provide a little more detail to the reader and allow it to store per-value context information so that it is not constantly jumping back and forth within the file unnecessarily.

### DIFF
--- a/examples/bufferedreader.c
+++ b/examples/bufferedreader.c
@@ -1,0 +1,807 @@
+/* vim: set sw=4 ts=4 et tw=80: */
+
+/****************************************************************************
+**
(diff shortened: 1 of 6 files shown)


sjlongland:
This follows on from https://github.com/intel/tinycbor/pull/208 as the changes would clash if they were kept separate.

PR #327, opened 2026-04-25
Add an example writer

sjlongland:
This provides an example implementation of the abstract writer interface using low-level file I/O as a demonstration.


sjlongland:
We can merge this into https://github.com/intel/tinycbor/pull/208 if desired, I created the separate branch expecting there'd be more, but it looks like nearly all the work was on the parser end.

PR #335, opened 2026-09-13
fix(cbortojson): keep strchr() result const for C23

suda-morris:
This PR fixed the following build warning:

```
/tinycbor/src/cbortojson.c: In function ‘escape_text_string’:                                                         
/tinycbor/src/cbortojson.c:344:21: error: initialization discards ‘const’ qualifier from pointer target type          
[-Werror=discarded-qualifiers]                                                                                                                                       
  344 | char *esc = c > 0 ? strchr(escapeChars, c) : NULL;                                                                                                           
      | ^                                                                                                                                                            
cc1: all warnings being treated as errors                                                                                                                            
configdep.c:684: in main: exec_process exited with error code: 1        
```

GCC in C23 mode gives strchr() the const of its argument, so assigning the result into a char * is -Wdiscarded-qualifiers. That becomes a build error under -Werror. esc is only used as a pointer into the const escapeChars table, so const is correct.


suda-morris:
@thiagomacieira PTAL

thiagomacieira (approved):

thiagomacieira:
Thank you!

